PR #30, opened 2023-01-26
BIP-164

solarcurvey:


Tritium-VLK (approved):
This sends BAL to the LM multisig `0xc38c5f97B34E175FFd35407fc91a937300E33860 ` so the maxis can bridge and deposit it. 

The JSON to deposit into Aave on Polygon can be more easily built after the funds have been transferred to the LM Multisig.

PR #37, opened 2023-02-01
Create BIP-169.json

solarcurvey:


Tritium-VLK:
Note: Executing this BIP will start a migration window where no gauge additions are possible for 14 days.  It should be carefully coordinated with other governance.

PR #47, opened 2023-02-03
Merge results from actions testing in fork back to master for final test in non-forked environment. 

Tritium-VLK:


Tritium-VLK:
See: https://github.com/stefanzweifel/git-auto-commit-action/issues/114

PR #55, opened 2023-02-08
Script to generate a list of actions for authorizing v3 factory pools.

Tritium-VLK:


Tritium-VLK:
See: https://forum.balancer.fi/t/bip-xxx-transfer-governance-control-on-polygon-arbitrum/4359

For end result of script outputs.

PR #90, opened 2023-02-26
Give signers gas + make Trit deployer

Tritium-VLK:
- Add tritium as a deployer
- Add signers with no deployer wallet to watch lists with .25 ETH topup amount (total of 3 non-deployer signers after Tritium)


Tritium-VLK:
```
Total eth requirements with all wallets full.

4 deployers * .5 = 2 eth

3 signers with .25 eth = .75 ETH

Total 2.75 ETH + at least 0.3 in the gas station for the next topup.
```

Tritium-VLK:
Nounce 86

PR #93, opened 2023-02-27
BIP-205: Enable relayers on gnosis chain and add other missing permissions 

Tritium-VLK:
This will find all action ids that should have permissions set based on the standard configuration here and generate a BIP to enable them.


Tritium-VLK:
https://forum.balancer.fi/t/bip-205-enable-relayers-on-gnosis-chain-and-add-other-missing-permissions/4451

PR #97, opened 2023-02-28
Revise full chain template with pool-recovery-manager and blab_ops msig

Tritium-VLK:
- Added Addresses for blabs_ops multisigs
- Cleaned up script and address book to better deal with addresses from monorepo
- Generated a new and updated template (config + outputs in docs dir)


solarcurvey (approved):
blabs multisigs all check out

PR #113, opened 2023-03-21
BIP-216: Add permissions for Batch Relayer V5 (second attempt)

Tritium-VLK:
Repeat attempt at BIP-216 after BIP-214 pulled V4 addresses due to the way targets were specified and an out of date address file.


Tritium-VLK:
https://forum.balancer.fi/t/bip-216-add-permissions-for-batch-relayer-v5-second-attempt/4534

PR #127, opened 2023-04-03
Sweep some tokens into the mimic mock collector on op

Tritium-VLK:
will send the entire collector balance of:

Premia, Pickle, Beets/wsteth, 3pool+beets+reth+bal

Debank says that should be worth about 120 bucks.


Tritium-VLK:
Second run is 5% of everything in the collector.

PR #143, opened 2023-04-18
BIP-254: Beethoven-X SP BAL airdrop

Xeonus:
- Airdrop for Beethoven-X worth $40k with BAL price at $7.19


ToterSunder:
иди нахуй

Tritium-VLK (approved):

PR #153, opened 2023-04-23
BIP-268: Grant ability to add 3rd party rewards on L2's with new gauges

solarcurvey:
https://forum.balancer.fi/t/bip-268-enable-non-bal-reward-tokens-on-new-l2-gauges/4689


Tritium-VLK (approved):
Generated by script, spotchecks look good, no other changes.

PR #158, opened 2023-04-26
BIP-273: Preset Update Request 1

santinomics:
https://forum.balancer.fi/t/bip-273-permissions-preset-update-request-1/4677


Tritium-VLK:
Went through file with karpatkey.  Everything seems to add up.  It's still not super easy to connect everything together and write review notes.  Provided some feedback to KPK to think about this for next time/continue to try to improve the process.

This looks ok to me/ seems to match the spec and docs provided by KPK in the BIP.

Tritium-VLK (approved):

PR #163, opened 2023-04-28
BIP-277 Enable Authorizer Wrapper

solarcurvey:


Tritium-VLK (approved):
Consulted with blabs who say this Authorizer wrapper has already undergone deep review.

PR #176, opened 2023-05-05
BIP-288 Disable Pool Factories

solarcurvey:


Tritium-VLK (approved):
Roles granted are roles revoked.
mainnet sim works

PR #191, opened 2023-05-17
[feat] - implement json merging algorithm POC

SHAKOTN:
## Done:
1. Walks through all dirs in `BIPs/` directory
2. Finds out what BIPs have `createdAt` within current week
3. Group them all by the chain and by msig address
4. In the end, collect all transactions from grouped BIPs and store to files in format `<chain_id>-<msig_addr>`

### DIFF
--- a/action-scripts/merge_pr_jsons.py
+++ b/action-scripts/merge_pr_jsons.py
@@ -61,6 +61,9 @@ def main():
     current_week = datetime.utcnow().strftime("%U")
     current_year = datetime.utcnow().year
     # get root directory of the project:
+    # To do this you need to go up 2 levels from the current file
+    # For instance, to get to the project root from: multisig-ops/action-scripts/merge_pr_jsons.py
+    # You need to jump up two steps with os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
     root_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
     bips_dir = os.path.join(root_dir, "BIPs")
     dir_with_target_bips = os.path.join(bips_dir, TARGET_DIR_WITH_BIPS)
--- a/action-scripts/merge_pr_jsons.py
+++ b/action-scripts/merge_pr_jsons.py
@@ -2,11 +2,8 @@ import json
 import os
 from collections import defaultdict
 from datetime import datetime
-from datetime import timedelta
-from datetime import timezone
 from json import JSONDecodeError
 from typing import Optional
-from typing import Tuple
 
 base_json = json.loads('''
 {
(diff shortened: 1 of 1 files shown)


action-scripts/merge_pr_jsons.py
```diff
@@ -23,43 +22,96 @@
...
-        safe = data["meta"]["createdFromSafeAddress"]
-        print(safe)
-        if safe not in txmap[chain_id].keys():
-            txmap[chain_id][safe] = []
-        txmap[chain_id][safe] = [*txmap[chain_id][safe], *data["transactions"]]
-        print(txmap)
+    current_week = datetime.utcnow().strftime("%U")
+    current_year = datetime.utcnow().year
+    # get root directory of the project:
+    root_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
Tritium-VLK, line 63: This is hard for me/anyone who isn't a real pythonista to understand.  Can you provide a comment?

Tritium-VLK: is this just pwd?  Why so many nested calls?

action-scripts/merge_pr_jsons.py
```diff
@@ -23,43 +22,96 @@
...
+    bips_dir = os.path.join(root_dir, "BIPs")
+    dir_with_target_bips = os.path.join(bips_dir, TARGET_DIR_WITH_BIPS)
```
Tritium-VLK, line 65: BIPs should be read from a path provided as an input for now (via env var or argv)

BIPs should be written to 00merged/YYYY-#WW  (or 00batched, there's one example there)

Tritium-VLK: the destination directory for the given run won't exist when the report runs.

SHAKOTN: That's clear. Gonna fix this

action-scripts/merge_pr_jsons.py
SHAKOTN: Sure. When you call `os.path.abspath(__file__)` that would be full absolute path to the executed python file. Like:
```
os.path.abspath(__file__) == ''opt/multisig-ops/action-scripts/merge_pr_jsons.py''
```

Then you just want to move up 2 times, for that you are using `os.path.dirname`, and you end up in `opt/multisig-ops/`, which is a root directory of your project

SHAKOTN: Comment added

action-scripts/merge_pr_jsons.py
SHAKOTN: Done

PR #219, opened 2023-05-29
Pull chain list from address book.

Tritium-VLK:


Tritium-VLK:
@SHAKOTN recently added zkevm, and the merge missed it.  Will be adding AVAX soon too, so better to pull from the addressbook, which gets updated when there are new chains.  Somehow this still isn't seeing the zkevm payloads in the 2023-W21 folder though.  Maybe you can investigate.

PR #220, opened 2023-05-30
Switch bb-e-usd to aave v3 for veBAL fee token.

Tritium-VLK:
Do not merge until after the next pay fees run.


Tritium-VLK:
Can someone else just triple check this token address and approve before Friday?

mikebmikeb (approved):

PR #236, opened 2023-06-02
Generate L2 Checkpointer Payloads

Tritium-VLK:
Create another payload to run after the governance payload to add new L2 gauges to the mainnet checkpointer.

### DIFF
--- a/action-scripts/merge_pr_jsons.py
+++ b/action-scripts/merge_pr_jsons.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from datetime import datetime
 from json import JSONDecodeError
 from typing import Optional
+from bal_addresses import AddrBook
 
 base_json = json.loads('''
 {
@@ -23,14 +24,6 @@ base_json = json.loads('''
 }
 ''')
 
-CHAIN_IDS = {
-    1: "mainnet",
-    3: "ropsten",
-    42161: "arbitrum",
-    100: "gnosis",
-    137: "polygon",
-    10: "optimism",
-}
 IGNORED_DIRECTORIES = ["examples", "rejected", "batched", "proposed"]
 # Place your BIPs json into this directory under BIPs/<TARGET_DIR_WITH_BIPS>
 TARGET_DIR_WITH_BIPS = "00merging"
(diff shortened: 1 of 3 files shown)


Tritium-VLK:
To test you can run on the 2023-W22 directory as it has some gauge adder v4 payloads that will trigger this.

action-scripts/merge_pr_jsons.py
```diff
@@ -125,11 +138,28 @@ def main():
             result["transactions"] = []
             for file in fs:
                 result["transactions"] += file["transactions"]
+                #  Check for gauge adds and generate checkpoint list
+
+                for tx in file["transactions"]:
+                    if tx["contractMethod"]["name"] == "addGauge":
+                        try:
+                            gauge_chain = tx["contractInputsValues"]["gaugeType"]
+                            if gauge_chain != "Ethereum":
```
SHAKOTN, line 147: What's the point of this check here?

action-scripts/merge_pr_jsons.py
```diff
@@ -51,16 +44,36 @@ def _parse_bip_json(file_path: str, chain: int) -> Optional[dict]:
...
                 return None
             # Check if chain id is the same as the one we are looking for
+            # TODO  This crashes if it finds JSON that is not a payload file.  Discovered by running on BIPs/
             if int(data["chainId"]) == int(chain):
                 return data
     except JSONDecodeError:
         return None
 
+def _write_checkpointer_json(output_file_path: str, gauges_by_chain: dict):
+    template_path = os.path.dirname(os.path.abspath(__file__)) + "/tx_builder_templates/l2_checkpointer_gauge_add.json"
```
SHAKOTN, line 54: If that's a template I recommend adding `template` to the filename :)

action-scripts/merge_pr_jsons.py
```diff
@@ -125,11 +138,28 @@ def main():
...
             for file in fs:
                 result["transactions"] += file["transactions"]
+                #  Check for gauge adds and generate checkpoint list
+
+                for tx in file["transactions"]:
+                    if tx["contractMethod"]["name"] == "addGauge":
+                        try:
+                            gauge_chain = tx["contractInputsValues"]["gaugeType"]
+                            if gauge_chain != "Ethereum":
+                                if gauge_chain not in gauge_lists_by_chain.keys():
```
SHAKOTN, line 148: You can simplify by doing this:
```python
from collections import defaultdict

gauge_lists_by_chain = defaultdict(list)
gauge_lists_by_chain[gauge_chain].append(tx["contractInputsValues"]["gauge"])
```

It will initialize empty lists for any new values so you don't have to do `if gauge_chain not in gauge_lists_by_chain.keys():` this check

action-scripts/merge_pr_jsons.py
```diff
@@ -125,11 +138,28 @@ def main():
...
+                                gauge_lists_by_chain[gauge_chain].append(tx["contractInputsValues"]["gauge"])
+                        except:
+                            print(f"Skipping checkpointer add for addGauge tx as it doesn't have expected inputs:\n---\n {tx['contractInputsValues']}")
+
             # Save the result to file
             file_name = f"{chain_id}-{safe_address}.json"
             file_path = os.path.join(dir_name_batched_full, file_name)
             with open(file_path, "w") as new_file:
                 json.dump(result, new_file, indent=2)
+    if gauge_lists_by_chain != {}:
```
SHAKOTN, line 159:
```suggestion
    if not gauge_lists_by_chain:
```

SHAKOTN (commented):
Some QOL suggestions

action-scripts/merge_pr_jsons.py
```diff
@@ -51,16 +44,37 @@ def _parse_bip_json(file_path: str, chain: int) -> Optional[dict]:
             if not isinstance(data, dict):
                 return None
             # Check if chain id is the same as the one we are looking for
+            # TODO  This crashes if it finds JSON that is not a payload file.  Discovered by running on BIPs/
             if int(data["chainId"]) == int(chain):
                 return data
     except JSONDecodeError:
         return None
```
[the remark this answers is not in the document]
Tritium-VLK: Which empty line, and was hoping you could fix that one at some point :)

action-scripts/merge_pr_jsons.py
Tritium-VLK: 1: We don't want to add Ethereum gauges to the L2 checkpointer.
2: We don't want to act unless it is an addGauge
3: There are older versions and maybe newer versions of the gaugeAdder that don't have gaugeType, and we don't want to add those.

action-scripts/merge_pr_jsons.py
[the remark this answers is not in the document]
SHAKOTN: Between line 52 and 53 :) Just add one more empty line. By default there should be 2 EL between 2 functions

action-scripts/merge_pr_jsons.py
SHAKOTN: Gotcha

SHAKOTN (approved):

PR #242, opened 2023-06-04
Fix boolean logic around when to create a checkpointer payload

Tritium-VLK:
Strange not crept in.


Tritium-VLK:
see #241 produced from this code/branch.

SHAKOTN (approved):

PR #258, opened 2023-06-12
feat: Gauge Add report v2

### LINKED ISSUE #205: Bounty:  Refactor Permissions report + addressbook + add more multisig payload autoverificiation
https://github.com/BalancerMaxis/multisig-ops/blob/main/action-scripts/brownie/scripts/get_gauge_mappings.py

Is a simple brownie based script that hops chains, pulls abis and contract data, while exploring json payload data to figure out details about gauge changes.

Over time Balancer has moved between different gauge types and methods of adding/removing gauges.  As a result, the script moved from a simple loop to a beast.

Task:

- Refactor this script to be cleaner, and to allow for additional handlers to be added without bloat.  It should be able to run on every commit/json payload and print/generate reports if there is something to be said.  Use bal_addresses.

- Add handlers to also cover grantRole and revokeRoll call.  In this case lists of bytes32 action ids are passed back and forth.  They can be looked up in the monorepo.    Speak to tritium for examples of code that does this in the Balancer Docs repo or also here.
   - At best, add additional data and functions to the addressbook to allow the reverse lookup of deployment/function by actionId.
   
- In the process, make use of bal addressbook and consider how to make it more enjoyable to work with.

Bounty:
Please come back with a proposal for this work.  As I see it there are 3 potential deliverables.

1: A gauge report script that is cleaner and makes it easier for me or others to add new handlers without breaking the whole thing/creating a huge mess.  Perhaps also some tests.

2: A handler for roles

3: Working towards v1 of the address book.

SHAKOTN:
## Context:

Initial gauge report script was written by @Tritium-VLK 
This PR is an attempt to rewrite and deliver a new and improved version for gauge report script

---
### Done
1. Introduce new script
2. Separate everything into smaller functions for better future extendability
3. Currently implement only gauge add mechanics

Partially Fixes #205


action-scripts/brownie/scripts/report_gauge_add.py
```diff
@@ -0,0 +1,165 @@
...
+ROOT_DIR = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+)
+GAUGE_ADD_METHODS = ['gauge', 'rootGauge']
+
+STYLE_MAINNET = "Mainnet"
+STYLE_SINGLE_RECIPIENT = "Single Recipient"
+STYLE_CHILD_CHAIN_STREAMER = "ChildChainStreamer"
+
+SELECTOR_MAP = {
```
Tritium-VLK, line 25: We should add zkevm and avalanche too.  I think in the case of new gauge adds with the adder v4 you can alwasy figure out which gauge it is by looking at the name of the add function called.

action-scripts/brownie/scripts/report_gauge_add.py
```diff
@@ -0,0 +1,165 @@
...
+        recipient = Contract(gauge.getRecipient())
+        escrow = Contract(recipient.getVotingEscrow())
+        pool_name, pool_symbol, pool_id, pool_address, a_factor, fee = get_pool_info(escrow.token())
+        style = STYLE_SINGLE_RECIPIENT
+    else:
+        (pool_name, pool_symbol, pool_id, pool_address, a_factor, fee) = get_pool_info(
+            gauge.lp_token())
+        style = STYLE_MAINNET
+
+    return {
```
Tritium-VLK, line 106: In gaugaddrv4 the chain name can be determined by the function call, but if it is not apparent, then it may not hurt to add that here as well, it does make the table wide.  Could ask the maxis and see what they think.

Tritium-VLK (commented):
Just a few comments.  Maybe I missed something in terms of picking up the calls.

but he gauge adder v4 takes Target and gaugeType (string) as inputs to addGauge.

The gauge type read function provides a list of strings, so assuming we only use v4 going forward maybe it makes more sense to focus that top level handler selection based on matching the gauge types.  Will make adding new ones easier as well.

Tritium-VLK (dismissed):
Looks better then what is there now.

A little unsure about where all these json files came from/if they are changing data that is already there.  Let's talk about it tomorrow, then we can merge and do any more changes on a fresh branch.

The other thing is, the old report also did kills and that has been removed now, so it would be good to get a kill report up soon to follow.

SHAKOTN:
@Tritium-VLK To answer shortly all those reports in my PR were modified by me so GH action has something to grind on

I will remove them today and open fresh pr

PR #318, opened 2023-06-30
Add payload reporting for transfers.

Tritium-VLK:
Not sure if this is the right way to integrate it with other reports but it makes sense.  Interesting we're just using brownie now ;)

Tested and used to generate reports for #316

### DIFF
--- a/.github/workflows/run_transfer_report.yaml
+++ b/.github/workflows/run_transfer_report.yaml
@@ -1,0 +1,54 @@
+name: Generate Transfers Payload Report
+
+on:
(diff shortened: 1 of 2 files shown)


SHAKOTN (commented):
Code looks good. Let me intergrate it with merge report, since it's handling all types of txs to post everything in one comment

Tritium-VLK:
Sounds good.  The last report that we need (once the rest are done) is the transactions with no report report, so it's good that we're running all the reports from a single handler that could eventually track such a thing.

PR #324, opened 2023-07-04
Bip tracking &  combine actions

Tritium-VLK:
- Integrated BIP and list index data into reporting and validation
- Merged some actions together, removed duplication
- Cleaned up actions so that all actions should exit ok if everything is good
- Added and Deleted a set of BIPs for testing


Tritium-VLK:
@SHAKOTN one question for your review.  The file name with spaces in it got encoded with extra info but didn't throw a report.  Any ideas?

.github/workflows/process_every_payload.yaml
```diff
@@ -0,0 +1,81 @@
...
+  COMMIT_SHA: ${{ github.event.pull_request.head.sha }}
+  WEB3_INFURA_PROJECT_ID: ${{ secrets.WEB3_INFURA_PROJECT_ID }}
+  ETHERSCAN_TOKEN: ${{ secrets.ETHERSCAN_TOKEN }}
+  ARBISCAN_TOKEN: ${{ secrets.ARBISCAN_TOKEN }}
+  GNOSISSCAN_TOKEN: ${{ secrets.GNOSISSCAN_TOKEN }}
+  POLYGONSCAN_TOKEN: ${{ secrets.POLYGONSCAN_TOKEN }}
+  OPTIMISMSCAN_TOKEN: ${{ secrets.OPTIMISMSCAN_TOKEN }}
+  ZKEVMSCAN_TOKEN: ${{ secrets.ZKEVMSCAN_TOKEN }}
+
+jobs:
```
SHAKOTN, line 20: I would advise against merging both actions into one. Failure in one of them will fail entire pipeline. Better to have them as one

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -124,6 +124,8 @@ def _parse_added_transaction(transaction: dict, **kwargs) -> Optional[dict]:
     pool_name, pool_symbol, pool_id, pool_address, a_factor, fee, style = _extract_pool(
         chain, gauge, gauge_selectors
     )
+    bip_number = transaction.get("bip_number")
```
SHAKOTN, line 127: Set some default values for better output: 
```suggestion
    bip_number = transaction.get("bip_number", "N/A")
```

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -124,6 +124,8 @@ def _parse_added_transaction(transaction: dict, **kwargs) -> Optional[dict]:
...
+    tx_index = transaction.get("tx_count")
```
SHAKOTN, line 128: Same here

action-scripts/brownie/scripts/script_utils.py
```diff
@@ -156,3 +157,50 @@ def merge_files(
...
         ])
     return merged_dict
+
+def get_bip_number(path: str) -> str:
+    """
+    Function that gets a BIP number from a file name by searching BIP-### pattterns.
+    Will throw a ValueError if exactly 1 result is not found, otherwise return a result like BIP-###
+    """
+    # Define the regular expression pattern
+    pattern = r"BIP-\d{3,}"
```
SHAKOTN, line 167: BIPs can be suffixed with letters, like `BIP-312A` and such

action-scripts/brownie/scripts/validate_bip.py
```diff
@@ -42,16 +43,37 @@ def validate_chain_specified(file: dict) -> Tuple[bool, str]:
...
     """
     chain = file.get('chainId')
-    if not chain or not isinstance(chain, str):
-        return False, "No chain specified or it is not a string"
+    chains = list(AddrBook.CHAIN_IDS_BY_NAME.values())
+    if int(chain) not in chains:
+        return False, f"No chain specified or is not found in known chain list: {chain} in {chains}"
     return True, ""
 
+def validate_txs_have_extra_data(file: dict) -> Tuple[bool, str]:
```
SHAKOTN, line 51: I can't understand why do we need to validate BIP payload that you've added yourself...

.github/workflows/process_every_payload.yaml
Tritium-VLK: I was thinking the opposite.  We don't want PRs to get merged unless all this stuff actually finishes in an expected way, and it all relates to each other.

Before lots of tests ended red/bad even if they didn't need to run.

We could use multiple jobs as well, but looking more into github actions there's a lot of benefits to keeping this stuff in one file.  Let's think about that more.

PR #330, opened 2023-07-05
Rework reporting and merge action flow

Tritium-VLK:


Tritium-VLK:
ok @SHAKOTN here are my tinkerings.  Pretty sure the on_workflow on the process action should trigger when the merge action runs but according to the docs it only runs if the workflow is merged to main.  So I think we gotta merge this to master, and then try running the payload merge from the main branch with the updated action in place.

SHAKOTN (approved):
Let's give this a try

PR #332, opened 2023-07-05
fix: Check chain exists and improve error message

Tritium-VLK:
Somehow this was throwing an error on some payloads before.  This both provides more details if an error is thrown, and also does more specific validation of the data.  (no rush to review/already merged into my action testing branch) #330


Tritium-VLK:
https://github.com/BalancerMaxis/multisig-ops/pull/330#issuecomment-1622084999

SHAKOTN (approved):
lgtm

PR #346, opened 2023-07-10
Rework reporting and merge flow

Tritium-VLK:
Failed first test.


Tritium-VLK:
ok @SHAKOTN I think this is all really working more (def a squash merge).

#352 is an example merge of 4269-W69 into this branch that I plan to close once we agree it is gud/working.

What a pain, but now I understand.

Tritium-VLK:
Spoke too soon.  So close.

Tritium-VLK:
ok #354

.github/workflows/process_every_payload.yaml
```diff
@@ -0,0 +1,14 @@
+name: Process BIP PR
```
SHAKOTN, line 1: why need this?

SHAKOTN (approved):

.github/workflows/process_every_payload.yaml
Tritium-VLK: The naame or the action.

The action is needed to trigger the reports when a PR is opened (under the on conditions listed in this file).

The Name probs isn't needed, but it makes it easier to look at.  What this does is runs on PR and does bip processing.

PR #360, opened 2023-07-12
Add BIP number validation

Tritium-VLK:


Tritium-VLK:
@SHAKOTN this works, need it, merging it.

PR #361, opened 2023-07-12
add additional validation for weekly directory placement

Tritium-VLK:
and create test file that should make it fail.


Tritium-VLK:
@SHAKOTN this works, blind merging it too :D

PR #390, opened 2023-07-24
remove BIP-375 from W29 execution, move BIP-366-B to W32 execution

Tritium-VLK:


Tritium-VLK:
Solarcurve, [24.07.23 17:40]
looks like we have the part about killing gauges in the wave 2 proposal going to execution this week?

Solarcurve, [24.07.23 17:40]
if so that's incorrect

Solarcurve, [24.07.23 17:41]
also there's an issue with ankrETH rate provider on arbitrum. integrations is workin with them but for now we'll hold off on adding that gauge. they are aware of this. i'll do a commit to remove it from the combined json

solarcurvey (approved):

PR #406, opened 2023-07-27
Update permissions builder to use more advanced bal addresses.

Tritium-VLK:


Tritium-VLK:
Worked on #405 where it is already merged.

Issue #435, opened 2023-08-02
report for call_data only txs


Tritium-VLK:
see #432 In this case even the no report report didn't catch it as there isn't any real ABI encoded I think.  These need to show up too such that every index has a report.

Tritium-VLK:
see PR for examples of how I used brownie to decode call data for further reporting.

Issue #434, opened 2023-08-02
Better reporting - reverted merged tx on base


Tritium-VLK:
https://github.com/BalancerMaxis/bal_addresses/pull/95

The above PR corrects a merged payload.  In the end there were grant_role type txs called to something other than the authorizer.

Learnings to implement.

1: Setup a validation to make sure all grantRole and grantRoles calls are to vault/Authorizer

2: You cant see the to: in the permissions report or the gauge report.  Maybe pack that in, although tit makes the table wide. Need reverse lookuped to in this case.

PR #436, opened 2023-08-02
update report to include more to info about the to of the TX

Tritium-VLK:
see #434 and #435


action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -132,12 +134,20 @@ def _parse_added_transaction(transaction: dict, **kwargs) -> Optional[dict]:
         chain, gauge, gauge_selectors
     )
 
+    addr = AddrBook("mainnet")
+    to = transaction['to']
+    to_name = addr.reversebook.get(to, "!!NOT-FOUND")
+    if to_name == "20230519-gauge-adder-v4/GaugeAdder":
```
SHAKOTN, line 140: No point in this check. Check line 108: function will do early return if `to` field is not a GaugeAdder

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -207,11 +226,12 @@ def _parse_permissions(transaction: dict, **kwargs) -> Optional[dict]:
     """
     Parse Permissions changes made to the authorizer
     """
+    if not transaction.get("contractInputsValues"):
```
SHAKOTN, line 229: Check here makes sense since `contractInputsValues` not checked below

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -234,14 +254,20 @@ def _parse_permissions(transaction: dict, **kwargs) -> Optional[dict]:
     if not isinstance(action_ids, list):
         print(f"Function {function} came up with {action_ids} which is not a valid list.")
         return
+    to = transaction['to']
+    to_name = addr.reversebook.get(to, "!!NOT-FOUND")
+    if to_name == "20210418-authorizer/Authorizer":
+        to_string = "Authorizer"
+    elif isinstance(to_name, str):
+        to_string = f"!!f{to_name}??"
```
SHAKOTN, line 262: Siggestion to ,ake it `NOT_FOUND` or something

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -188,13 +200,20 @@ def _parse_removed_transaction(transaction: dict, **kwargs) -> Optional[dict]:
     pool_name, pool_symbol, pool_id, pool_address, a_factor, fee, style = _extract_pool(
         chain, gauge, gauge_selectors
     )
+    addr =AddrBook("mainnet")
+    to = transaction['to']
+    to_name = addr.reversebook.get(to, "!!NOT-FOUND")
+    if to_name == "20221124-authorizer-adaptor-entrypoint/AuthorizerAdaptorEntrypoint":
+        to_string = "AAEntrypoint"
+    elif isinstance(to_name, str):
+        to_string = f"!!f{to_name}??"
```
SHAKOTN, line 209: Siggestion to ,ake it `NOT_FOUND` or something

action-scripts/brownie/scripts/report_gauges.py
Tritium-VLK: yeah but which GaugeAdder.  it's important it's the right one, and sometimes we want to know which one it is. That's why I added to.  People notice stuff this way.

Tritium-VLK: I'll update line 108 to work for any gauge addr and then leave this.  In the end part of it is about a more consistent report format.

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -151,6 +161,8 @@ def _parse_removed_transaction(transaction: dict, **kwargs) -> Optional[dict]:
     """
     Parse a gauge remover transaction and return a dict with parsed data.
     """
+    if not transaction.get("contractInputsValues"):
```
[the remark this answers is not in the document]
Tritium-VLK: Where?

The next thing i see is on line 178 where it will crash cuz there is no get.

action-scripts/brownie/scripts/report_gauges.py
Tritium-VLK: In this case it is found, and has a name, it's just not the correct one. If it is really NOT-FOUND it is already set to NOT-FOUND by line 258.

action-scripts/brownie/scripts/report_gauges.py
Tritium-VLK: same

action-scripts/brownie/scripts/report_gauges.py
Tritium-VLK: NO that just makes things more fragile. W ill leave the code as it is, but I think this is ok like this.  We're just normalising the tables to look similar.

SHAKOTN (approved):

PR #528, opened 2023-09-27
BIP-442: Permissions Preset Update Request #2

xtom4s:
Json payload required for the Preset Update #2


BIPs/2023-W39/BIP-442-permission-preset2.json
```diff
@@ -0,0 +1,262 @@
+{
+  "version": "1.0",
+  "chainId": "1",
```
Tritium-VLK, line 3: Note that the safe address is missing here and it is no where specified in the BIP to which safe this payload should be loaded.

Could you please update the payload as well as including this information in the form post somewhere?

Tritium-VLK (changes requested):

Tritium-VLK:
Tenderly against eth:0x0EFcCBb9E2C09Ea29551879bd9Da32362b32fc89 

https://dashboard.tenderly.co/public/safe/safe-apps/simulator/2f82731d-ae49-47d4-9c83-89c82259561a

BIPs/2023-W39/BIP-442-permission-preset2.json
xtom4s: Done

PR #566, opened 2023-10-16
Moved the custom brownie branch to the BalancerMaxis repo 

Tritium-VLK:
Using that so we can all handle rpc problems/etc.


Tritium-VLK:
Take note of how this is wired up ser.

PR #594, opened 2023-10-30
Add missing W43 core pools

Tritium-VLK:
@gosuto-inzasheru @Xeonus remember to add these alongside the payloads in gauge add prs.


Xeonus (approved):
All correct from my side. Matches Gauge add requests as per W44 round

PR #608, opened 2023-11-07
zkEVM core pool updates

Zen-Maxi:
Add wstETH/weth, weth/rETH, and wstETH, rETH to core pool list.


SHAKOTN (approved):
Looks good to me, checked all pools manually. Need to be confirmed by other maxis

Tritium-VLK (approved):
These were all core pools before we changed the system, zkEVM wasn't added at the time.  Seems ok to do without a BIP.

PR #624, opened 2023-11-11
Remove R pool from core pools

Tritium-VLK:
It was just hacked, probably shouldn't pay vote incentives for it.


Zen-Maxi (approved):
Hope R can recover. Will communicate this to them as wlel.

PR #636, opened 2023-11-21
chore: bump `web3` dep

gosuto-inzasheru:
this adds support for python 3.11 as per https://github.com/ethereum/web3.py/pull/2699

### DIFF
--- a/action-scripts/requirements-actions.txt
+++ b/action-scripts/requirements-actions.txt
@@ -2,7 +2,7 @@ munch==4.0.0
 dotmap==1.3.30
 pathlib==1.0.1
 pandas==1.5.3
-web3==5.31.3
+web3==6.11.3
 tabulate==0.9.0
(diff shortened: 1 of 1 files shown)


Tritium-VLK (commented):
I think the issue is that it changes the some functions to checksum addresses from camel case to _ and the old format is used all over.  Did you test the current actions with the new requirements?

gosuto-inzasheru:
not sure i follow your comments.

im happy to not make changes, but can you specify the python version that was used for this current config then? 3.11 fails due to the mentioned bug in `web3`

gosuto-inzasheru:
ah i see now that the action scripts are running in ci, and there they run in 3.9: https://github.com/BalancerMaxis/multisig-ops/blob/main/.github/workflows/run_auth_gen.yaml

i ran them locally, but then there is no python version defined anywhere and my env defaulted to 3.11.

set my python to 3.9 locally and all works without this fix being needed

PR #648, opened 2023-11-24
fix: add `bal-addresses` to pyproject

gosuto-inzasheru:
fixes some issues that prevent installing the `tools/python/brownie` env out of the box using `poetry`


Tritium-VLK (commented):
why are you adding ape_safe.py here?  It was being pulled from imports before right?

Tritium-VLK (approved):

PR #652, opened 2023-11-25
feat: determine chain by gauge contract name

gosuto-inzasheru:
previously `chain` was determined by the function selectors present in the gauge contract, which for avalanche for example isnt conclusive. this adds logic that checks the gauge name first to determine the child chain


Tritium-VLK:
To test these, there's a ton of payload files in BIPs/4269-W69.

To test these, I tend to find something that has a lot of what we are changing and copy it in and/or move it around.

We do regular gauge kill cleaning bips that tend to have a lot of gauges on a lot of changes.

Here is an example:
https://forum.balancer.fi/t/bip-326-cleanup-unused-gauges/4861

Maybe see if you can find gauge kills from a bunch of chains, create a gaugeKill folder in 4269-W69, inlcude them all in this PR and make sure the reports work as a final test?

gosuto-inzasheru:
more testing in https://github.com/BalancerMaxis/multisig-ops/pull/654 (pr can be discarded, is just for running tests in ci)

Tritium-VLK (approved):

PR #654, opened 2023-11-26
chore: more gauge name reporting testing

gosuto-inzasheru:
running the report over some old gauge kills as well to stress test it


gosuto-inzasheru:
@Tritium-VLK testing for #652, lgtm

PR #656, opened 2023-11-27
ci: cache compilers for `brownie` actions

gosuto-inzasheru:
introduces caching for workflows that rely on `brownie`. prevents having to download all `solc` and `vyper` binaries at every run.

via https://github.com/brownie-mix/github-actions-mix/blob/0bfdd5fab8ece6ab13d757173c9e8f68501ee998/.github/workflows/main.yaml#L22-L28

### DIFF
--- a/.github/workflows/merge_json.yaml
+++ b/.github/workflows/merge_json.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       runWeek:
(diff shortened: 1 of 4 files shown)


Tritium-VLK (approved):
Looks like a good change if you tested everything. I'll be OoO so if you merge it and stuff starts breaking make sure you are both around to roll it back/notice.

SHAKOTN (approved):
LGTM

PR #667, opened 2023-11-29
FIX: Add weETH-WETH as core pool

Xeonus:
- BIP-492 should have had core-pool designation but this was missed. This PR fixes the config to add that pool


Zen-Maxi (approved):
lgtm

PR #673, opened 2023-12-06
BIP-501: Re-Enable the THX/USDC Gauge with 2% cap [Polygon]

Xeonus:
- enalbe THX/USDC 2% gauge on Polygon
- Tenderly-sim: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/5dcff9c6-31eb-4f92-b53c-2c191769e420


gosuto-inzasheru (commented):
payload checks out, just wondering if we should not also disable the old thx gauge?

PR #682, opened 2023-12-13
BIP-506: Enable 80MIMO/20WETH Gauges with a 2% cap

Xeonus:
- add MIMO-WETH gauges with a 2% cap to Mainnet and Polygon
- Tenderly-sim: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/9934a641-667d-4f88-8a98-f5ca0ce4f315


gosuto-inzasheru (approved):
both are capped, gauge and pool addresses check out. no rate providers

PR #706, opened 2023-12-21
Core pool fixes

Tritium-VLK:
On matic 2 old and/or no liquidty pools were listed as core, and the pools that had liquidty and fees were not.  This swaps them out.  

Also @SHAKOTN , `0xb9debddf1d894c79d2b2d09f819ff9b856fca55200000000000000000000062a` rETH/RPL has decent TVL, should have decent fees and get a brib.

Other than that things look ok for this round.  Should figure out 80mevETH/20fold soonish.  It is not empty should have some fees, they may be low but they are not 0.  In other cases low/non-0 fees are shown so something is wrong.  If this pool started earning we would miss it.  I can make a ticket for that if you want, just ping me. 

We already discussed qETH and weETH.  Have pricing problems there, will assume that both tokens are worth 1 ETH for pricing for this round.

https://github.com/BalancerMaxis/protocol_fee_allocator/blob/main/fee_allocator/allocations/incentives_2023-12-07_2023-12-21.csv


config/core_pools.json
```diff
@@ -26,11 +26,11 @@
       "0x05ff47afada98a98982113758878f9a8b9fdda0a000000000000000000000645": "weETH/rETH"
     },
     "polygon": {
-      "0xf0ad209e2e969eaaa8c882aac71f02d8a047d5c2000200000000000000000b49": "ECLP-WMATIC-stMATIC",
+      "0x8159462d255c1d24915cb51ec361f700174cd99400000000000000000000075d": "ECLP-WMATIC-stMATIC",
```
gosuto-inzasheru, line 29: the old one has $5.7m tvl, the new one only $1.1m?

gosuto-inzasheru (changes requested):
dont really understand why the eclp-wmatic-stmatic has to be swapped out?

config/core_pools.json
SHAKOTN: The new one is not gyro pool as well?

gosuto-inzasheru: aaah it should be replacing the non gyro on L31

gosuto-inzasheru:
sorry now the diff is all weird because the file was badly linted, but should be fixed now

Tritium-VLK:
Ok sorry for the confusion.  This looks good now.  merging.

SHAKOTN (approved):
2 new poly pools look good.

gosuto-inzasheru (approved):

PR #717, opened 2024-01-02
[BIP-525] Business Development Funding Proposal Q1-2024 and Retroactive Q4-2023 Funding

gosuto-inzasheru:
forum: https://forum.balancer.fi/t/bip-xxx-business-development-funding-proposal-q1-2024-and-retroactive-q4-2023-funding/5481/2
sim: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/6e085de7-5cdb-480b-9750-024b4de871b0/


gosuto-inzasheru:
`0xC7E84373FC63A17B5B22EBaF86219141B630cD7a` is not caught by the report atm, needs https://github.com/BalancerMaxis/bal_addresses/pull/170/files merged

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -506,7 +506,11 @@ def parse_no_reports_report(
             contractMethod = transaction.get("contractMethod", {})
             no_reports.append(
                 {
-                    "fx_name": contractMethod.get("name", "!!N/A!!"),
+                    "fx_name": (
```
Tritium-VLK, line 509: What is this change about @gosuto-inzasheru.  I think that if/else was there for a reason and it's fragile.  Can we abstract this to another PR?

Tritium-VLK (changes requested):

action-scripts/brownie/scripts/report_gauges.py
gosuto-inzasheru: i still think it is an issue that these ci reports crash silently, it should be obvious that they didnt complete their run.

Tritium-VLK (approved):

Tritium-VLK (approved):

BIPs/2024-W2/BIP-525.json
```diff
@@ -0,0 +1,56 @@
...
+        "name": "transfer",
+        "payable": false
+      },
+      "contractInputsValues": {
+        "to": "0xC7E84373FC63A17B5B22EBaF86219141B630cD7a",
+        "value": "20000000000"
+      }
+    },
+    {
+      "to": "0xC7E84373FC63A17B5B22EBaF86219141B630cD7a",
```
Tritium-VLK, line 49: This does not seem like a complete transaction. Is it trying to be a transfer?

action-scripts/brownie/scripts/report_gauges.py
Tritium-VLK: @gosuto-inzasheru I think the issue is that the payload is wrong and the second tx is a malformed transfer.   I guess it makes sense to have it finish and post an error or something, but maybe we should be louder about the fact that something is wrong with one of the transactions.  Can we abstract this into another PR and think about how to handle it.  There are some other calldata only use_cases we should think about/see how they interact.

Tritium-VLK (changes requested):

Tritium-VLK (approved):

BIPs/2024-W2/BIP-525.json
Tritium-VLK: Spoke about this on a call.  THis is an ETH transfer. Created ticket to further improve reporting.

PR #723, opened 2024-01-04
BIP-527-2eur-Gauges

Zen-Maxi:
Parts 1 and grant / revoke roles for DAO Multisig to unkill gauges. Part 2 unkills the 2eur gauge 0x16289F675Ca54312a8fCF99341e7439982888077. Part 4 kills the correct 2eur gauge 0x790DE8ABE859f399023BCe73B5FE5C4870cD816A

Simulation: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/72295c0c-8a64-457f-8319-3f9fdc3a4580

### DIFF
--- a/BIPs/2024-W2/BIP-527.json
+++ b/BIPs/2024-W2/BIP-527.json
@@ -1,0 +1,83 @@
+{
+  "version": "1.0",
+  "chainId": "1",
+  "createdAt": 1704378156791,
+  "meta": {
+    "name": "Transactions Batch",
+    "description": "",
+    "txBuilderVersion": "1.16.3",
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
why is it necessary to `grantRole` for `unkillGauge()` twice?

gosuto-inzasheru (changes requested):
i think the second call to the authorizer should be to revoke said role

Zen-Maxi:
Good catch, rushing this through i did not change the second procedure to revoke. It is fixed now.

Xeonus:
Verified root gauges again for correctness. Now proper kill will be executed and unkill will revert wrong kill of 2eur (PAL)

gosuto-inzasheru (approved):

PR #727, opened 2024-01-06
feat: Add reports for incentive payloads.

Tritium-VLK:
testing:
- Add bribe payloads to test directory
- twiddle other payloads to make sure nothing broke


SHAKOTN (approved):

Tritium-VLK:
> how come some of the pool ids come back as `Not Found`?

The API does not return pool id for aura guages.

PR #737, opened 2024-01-10
BIP-529: Enable sFRAX-sfrxETH gauge [Optimism]

Xeonus:
- Enable sFRAX-sfrxETH gauge on OP
- Add pool to core pool list
- Tenderly-sim: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/9b54a936-e281-497c-8a46-5bedb4b2607e


config/core_pools.json
```diff
@@ -67,6 +67,7 @@
     "0xb26f0e66317846bd5fe0cbaa1d269f0efeb05c9600000000000000000000001e": "USDC-USDT"
   },
   "optimism": {
-    "0xa71021492a3966eec735ed1b505afa097c7cfe6f00000000000000000000010d": "frxETH/sfrxETH"
+    "0xa71021492a3966eec735ed1b505afa097c7cfe6f00000000000000000000010d": "frxETH/sfrxETH",
+    "0x2feb76966459d7841fa8a7ed0aa4bf574d6111bf00020000000000000000011d": "sFfrax-sfrxETH"
```
gosuto-inzasheru, line 71:
```suggestion
    "0x2feb76966459d7841fa8a7ed0aa4bf574d6111bf00020000000000000000011d": "sFRAX-sfrxETH"
```

typo in the label. however, more importantly, i thought we didnt manage core pools for beethoven?

gosuto-inzasheru (commented):
gauge address checks out.

not sure about the core pool thing, see inline comment

config/core_pools.json
Xeonus: @Tritium-VLK - how shall we proceed? Having a reference in our repo for all core pools (independent of its management) is beneficial in my opinion.

Tritium-VLK: We're moving to automatic core pool selection soon anyway so it will be less relevent.  I don't think we really need this, but I also think it is fine.  Better to remember to always add core pools to the list than.

Tritium-VLK (approved):

PR #801, opened 2024-02-16
Reroute migrated wstETH pool to new one with no yields

Tritium-VLK:
I think this may be one of the reasons the difference is so high on polygon, and this can drive deposits to the new pool.


Tritium-VLK:
Nevermind.  Missed some } and I don't think it is enough AUM to matter, but maybe we should try standing up a pool there.  IT would be easy and get volume.

PR #805, opened 2024-02-20
Revert auto corepools

Tritium-VLK:
Still problems in the logic.  Too many changes.  Need to get the system stable, let's manage static config for another month or 2.


Xeonus (approved):

Xeonus:
> why not use the white and blacklist to make manual changes?

Has to be a complete rollback as per Trit

PR #808, opened 2024-02-20
add missing core pool (weETH/rETH)

Tritium-VLK:


gosuto-inzasheru:
@Xeonus i found a solution for the fee exemption, but could you also elaborate on "not a proper core pool"?

PR #809, opened 2024-02-21
Schedule DAO Multsig gap

Tritium-VLK:
Move hypernative from W8 to W10
Leave readme in W9 asking to reschedule DAO txs to W10.

Tritium's taking a week off


Xeonus (approved):
As discussed, lgtm

PR #815, opened 2024-02-22
Add missing core pool

Xeonus:
- add sDOLA-DOLA stableswap pool to core pool list


Tritium-VLK (approved):
Confirmed this is sDOLA/DOLA on mainnet.

PR #827, opened 2024-03-02
HH_Rebates_and_Debt

Zen-Maxi:
Explains the HH refund standard procedure and explains the net debt owed to Hidden Hand due to miscalculated refund amounts in December 2023 and January 2024. Both Daos agreed to pay off this debt with future refunds until the net value is covered. Then refunds will continue as intended.


Xeonus (approved):
Thanks for documenting the state of HH rebates

PR #833, opened 2024-03-07
STIP period ending 2024-03-07

Xeonus:
- STIP run for 2024-03-07
- [Validation /comparison with previous run](https://docs.google.com/spreadsheets/d/1-KFc1vEFiV-Yw7BPFJzSDhI5WbgOtvCeJVC_IDCyeqw/edit#gid=1678132273)
- [Tenderly Sim](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/8991ffdc-61a5-497e-947c-447510fd9c94)

TODO: SIGN SAFE [HERE]()
TODO: SIGN DEN [HERE]()


Tritium-VLK:
Note that I have modified the payloads as follows:

- I removed both of the rETH pools from the program and the $ARB transfer in the original json
- I created a new json which assigned that full amount to the new rETH/WETH pool
- I independently did the same math on this sheet (see colored colums): https://docs.google.com/spreadsheets/d/1cayoL7FAhvGxiDeOKpH9LrHE6c-NKgnEeRKsNEN1egg/edit#gid=1517801612
- I then modified the actual table (another tab there) and dumped out another csv with final results.

Please review carefully, diffs should be clear and line up and amounts should match between the sheet and the payloads.

Total spend should add up to what's left in the safe, at best to the wei.  Anything left over we should send back to arbi and/or add to something as a last hooorah.

Xeonus:
- reroute new rETH gauge confirmed: 0x8ba2D53F34159C5C5e7add60B56C7dE3BBc1DA68
- reroute amount combined: 4891.63 per period confirmed
- reroute not in main payload confirmed
- checked amounts and they add up as per sheet
- redeem all plans method present in both payloads

Tritium-VLK (approved):

PR #839, opened 2024-03-12
Improve reporting to cover more use cases and make more numbers more human readable

Tritium-VLK:
- cover eth transfers and situations where the values in contractinputsvalues is not a list.
- make better use of a similar function for handling tx-builder json formatted list strings
- denominate stuff that looks like token amounts with something more human readable


action-scripts/brownie/scripts/script_utils.py
```diff
@@ -432,25 +432,28 @@ def prettify_contract_inputs_values(chain: str, contracts_inputs_values: dict) -
...
+
+            ## Reverse resolve authorizor roles
+            elif "role" in key.lower():
                 outputs[key].append(
                     f"{value} ({perm.paths_by_action_id.get(value, 'N/A')}) "
                 )
+            ## If it looks like an amount provide 1e18 denomination (hard to understand what decimals will so make it clear.
+            elif "value" in key.lower() or "amount" in key.lower():
+                try:
+                    outputs[key].append(f"{value} / 1e18 = {int(value)/1e18}")
```
gosuto-inzasheru, line 451: are you sure we can always assume 18 decimals here?

gosuto-inzasheru (commented):
i dont see how eth transfers are now "covered" or handled differently? they were already covered afaict?

action-scripts/brownie/scripts/script_utils.py
Tritium-VLK: @gosuto-inzasheru No we can't...  The problem is, in a generalized report there is no way to know what decimals to assume.  That's why it clearly states over 1/e18 in the string so that someone can know what is going on and make a determination.  We could leave this out if you don't think it helps, but I often want for it.  Any thoughts?  Also it's not working/ trying to figure out this logic better if you have ideas there :)

In other news, we are running on 0.9.0 which is the new ecosystem-balances release/that looks good after some tweaks.

Tritium-VLK: Ok so I thought about this more.  When we know decimals or can know it I am using it.  When not will show both 18 and 6 decimals to make it clear that it is unclear that the actual answer is, but still be helpful.

gosuto-inzasheru (commented):
adds more human readable output for amounts expressed in wei

i left some minor comments

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -203,10 +212,12 @@ def _parse_set_receipient_list(transaction: dict, **kwargs) -> Optional[dict]:
     return {
         "function": "setRecipientList",
         "chain": chainbook.chain,
+        "injector": f"{to_address}({chainbook.reversebook.get(to_address, 'Not Found')})",
+        "symbol": symbol,
         "gaugeList": json.dumps(pretty_gauges, indent=1),
-        "amounts_per_period": json.dumps(pretty_amounts, indent=1),
```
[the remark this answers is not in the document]
Tritium-VLK: prettify_int_amounts (line 201) returns a prettified string, not json to be dumped.

Tritium-VLK (commented):
ok @gosuto-inzasheru I think I resolved all your suggestions/questions.

Tritium-VLK:
So @gosuto-inzasheru think we are good to merge this.  can you approve?

Tritium-VLK:
Lol oops.  Forgot to squash merge...

PR #841, opened 2024-03-14
[BIP-560] Enable Single Recipient Gauge for GEM/ETH on [Ethereum]

Xeonus:
- enable ve80/20 GEM/ETH single recipient gauge on mainnet
- recipient is a msig that approves spending on the reward manager contract to distribute to vlGEM


gosuto-inzasheru:
report script currently breaks on the recipient, which is their multisig, not having the `getVotingEscrow` attribute (https://github.com/BalancerMaxis/multisig-ops/actions/runs/8276693450/job/22645692288#step:7:227). thus being unable to programmatically retrieve the address of the `veGEM` token

Xeonus:
withhold and move to following week

Xeonus:
@Tritium-VLK @gosuto-inzasheru This is the answer from the proposer: https://forum.balancer.fi/t/bip-xxx-enable-single-recipient-gauge-for-gem-eth-on-ethereum/5608/3

`Basically MSig receives yield from the SingleRecipientGauge weekly and it will be queued on the vlToken contract to be distributed.
Later on we can see for an onchain permisionless system`

I think we can move forward this week unless there are objections

gosuto-inzasheru (approved):
yield is being distributed manually (permissioned) via a multisig

PR #849, opened 2024-03-20
BIP-569: BD funding Q2 2024

Xeonus:
- fund BD SP for Q2 2024
- 61,000 USDC and 6,342 BAL


gosuto-inzasheru (approved):
matches proposal as on the forum

PR #857, opened 2024-03-21
ci: install dependabot for gh actions

gosuto-inzasheru:
some actions are starting to give deprecation warnings


gosuto-inzasheru:
eg
Screenshot 2024-03-21 at 15 43 22

Tritium-VLK (approved):

PR #867, opened 2024-03-26
Handle no info single recipient gauge

Tritium-VLK:


action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -109,17 +109,30 @@ def _extract_pool(
...
+                pool_name,
+                pool_symbol,
+                pool_id,
+                pool_address,
+                a_factor,
+                fee,
+                tokens,
+                rate_providers,
+            ) = get_pool_info(escrow.token())
+        except:
```
gosuto-inzasheru, line 124: maybe make this except a bit more explicit, in case not having `getVotingEscrow` is not the problem:
```suggestion
        except AttributeError:
```

Tritium-VLK:
Included in #864

Xeonus (approved):

PR #853, opened 2024-03-21
BIP-575: Enable stataArbUSDT/stataArbUSDCn gauge [Arbitrum]

Xeonus:
- enable E-CLP gauge for stataArbUSDT/stataArbUSDCn gauge on Arbitrum
- rate providers are implemented with the ERC4626 standard
- rate provider review: https://github.com/balancer/code-review/pull/33
- Forum post: https://forum.balancer.fi/t/bip-575-enable-stataarbusdt-stataarbusdcn-gauge-arbitrum/5643

PR #878, opened 2024-03-29
Additional logic to handle gyro rehype rate providers which are provided under different read functions.

Tritium-VLK:
See #853 for an example run, this branch was merged into that PR.

### DIFF
--- a/action-scripts/brownie/scripts/script_utils.py
+++ b/action-scripts/brownie/scripts/script_utils.py
@@ -132,8 +132,8 @@ def get_pool_info(
     if len(rate_providers) == 0:
         try:
             rehype_pool = Contract(pool_address)
-            rate_providers[0] = rehype_pool.rateProvider0()
-            rate_providers[1] = rehype_pool.rateProvider1()
+            rate_providers.append(rehype_pool.rateProvider0())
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
am i missing something? https://github.com/BalancerMaxis/multisig-ops/pull/853#issuecomment-2027153375 still shows rate provider as empty

gosuto-inzasheru:
it is because of the way you add the rate provider to the list:
```
>>> li = []
>>> li[0] = 'a'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: list assignment index out of range
```

should just use `.append` i think

gosuto-inzasheru:
rate providers column is still empty.

suggesting to test this locally first and not here in prod

Tritium-VLK:
In general I would really prefer complete reports on the final bathced payload, so kind of trying to move a little quick to get these dealt with during the 5 days we have (most of which is a long weekend).  That being siad you are right :)

Tritium-VLK:
Ok @gosuto-inzasheru .  tested in dev env and in prod and looks good. agreed?

https://github.com/BalancerMaxis/multisig-ops/pull/853#issuecomment-2027195596

PR #889, opened 2024-04-04
Remove RETH rerouting

Tritium-VLK:
Gauges have been killed, this config is now causing the fee allocator to revert as demonstrated here:

https://github.com/BalancerMaxis/protocol_fee_allocator/actions/runs/8554837599/job/23440975596


Xeonus (approved):
Removal of old / killed IDs for both core pool json and rerouting config confirmed.

Issue #890, opened 2024-04-04
fix: ci report prints amounts twice


gosuto-inzasheru:
amounts are now printed twice, only raw and once with the decimal modifier. eg see https://github.com/BalancerMaxis/multisig-ops/pull/882#issuecomment-2034663313
```
"amount": [                                                                   
  [                                                                           
    "raw:275000000000000000000000, 18 decimals:275000.0, 6 decimals: 2.75e+17"
  ],                                                                          
  [                                                                           
    "275000000000000000000000"                                                
  ]                                                                           
]  
```

gosuto-inzasheru:
closed by https://github.com/BalancerMaxis/multisig-ops/pull/899

PR #899, opened 2024-04-16
Fix:  report ci: Avoid scenitific numbers on human readable wei converstion

Tritium-VLK:
Solves #890


gosuto-inzasheru (changes requested):
point of issue #890 is that now the amount is shown twice:

Screenshot 2024-04-16 at 14 12 18

the original wei (second list) should be removed imo

also i dont think casting to int is a smart idea here, what if there are actual decimals? eg usdc123,45?

Tritium-VLK:
Good thoughts, how would you recommend casting these?  I kind of actually think it was ok that if it gets super large it goes to scientific notation.  Trying to figure out why the raw amount is duplicated.

gosuto-inzasheru:
fix in cdabfe2

suggesting to use `Decimal` instead, this will prettify based on whether the decimal is significant or not and prevent floating precision loss

gosuto-inzasheru (approved):

gosuto-inzasheru:
i cant request review from you @Tritium-VLK, but just merge as a review i guess :)

gosuto-inzasheru (approved):

gosuto-inzasheru:
oh lol was on auto merge i think

PR #903, opened 2024-04-18
BIP-585: CSP V6

Tritium-VLK:


gosuto-inzasheru (commented):
not sure if i am fully capable of reviewing this properly.

ultimately it boils down to the config file used for `gen_add_permissions_payload.py`: https://github.com/BalancerMaxis/multisig-ops/blob/a50c6a2d9d17d7630ec3e7b9476c4c136b2b53d1/BIPs/2024-W16/BIP-585/inputs.json

some questions around it:
- where does the list of functions (`"function_caller_map"`) come from? i dont see all those functions on the `composable-stable-pool-v6` deployment (`0x5B42eC6D40f7B7965BE5308c70e2603c0281C1E9`)
- why are some of the callers to be authorised in a list and some arent?
- i understand cow and 1inch (should) have permissions to adjust the swap fee? is it true they (will) only do that on ethereum, and ethereum/polygon/arbitrum respectively?

~~then also, i cant find the mentioned `chainname_CSP5_disable.json` anywhere?~~ ah they are actually called `disable_chainname.json`

Tritium-VLK:
1: The only non CSPv5 based permissions are the CSPv5 based disabled permissions. This is a template that we use a copy of last pool changes.  The srcipt will not apply permissions that do not exist, so it could be the case that there are some functions for weighted pools in the input.json that do not make it into the resulting CSV.  What specific functions are you seeing in the input.json that are not in the cspv5 description.  I made the config by pulling toghether past config files from [BIP-368]https://forum.balancer.fi/t/bip-369-grant-cspv5-swap-fee-permissions-to-cowswap-1inch-labs-solver/5003 and [BIP-353](https://forum.balancer.fi/t/bip-353-grant-permissions-for-composable-stable-pool-factory-v5/4974). Can you be more specific about where you see a discrepancy?

2: can you provide more details.  If you mean in the input json it can accept either lists or strings for one or more callers.


3: I asked about this in the integrations slack and got no answers.  In the end this is hwat we have always setup in the past, so it seems the thing to do to move the permisisons from CSPv5 over.

gosuto-inzasheru (approved):
- ok great i hadnt found bip353 yet. i compared the proposed input file with [the one from bip353](https://github.com/BalancerMaxis/multisig-ops/blob/main/BIPs/2023/2023-W28/BIP-353/inputs.json). lgtm.
- for example why have a list here (`"setSwapFeePercentage(uint256)": ["multisigs/lm"],`) but then a single string here: `"startAmplificationParameterUpdate(uint256,uint256)": "multisigs/lm",`. if it accepts both then i guess it functionally doenst matter
- makes sense

Tritium-VLK:
> * ok great i hadnt found bip353 yet. i compared the proposed input file with [the one from bip353](https://github.com/BalancerMaxis/multisig-ops/blob/main/BIPs/2023/2023-W28/BIP-353/inputs.json). lgtm.
> * for example why have a list here (`"setSwapFeePercentage(uint256)": ["multisigs/lm"],`) but then a single string here: `"startAmplificationParameterUpdate(uint256,uint256)": "multisigs/lm",`. if it accepts both then i guess it functionally doenst matter
> * makes sense

My guess is that at some point there was more than one member and it was a list,t hen it was removed but the list brackets were not.  I copied this config from prior bips, so it's kind of just is.

PR #904, opened 2024-04-18
BIP-591: Myso Call Option Trial

Xeonus:
- Transfer 25'000 BAL to LM Multisig
- Add template transactions to create vault, transfer BAL to vault and create quote. Needs to be adjusted at time of execution with latest parameter set
- Proposal: https://forum.balancer.fi/t/bip-591-generating-usdc-revenue-for-the-treasury-via-bal-covered-call-lending-on-myso/5665


gosuto-inzasheru (dismissed):
- txs lgtm, just need to be placed in the right folder month still i guess
- the failing tx is due to the vault not being deployed yet
- onchain quote review (with passed values in inline comments):

```solidity
struct GeneralQuoteInfo {
    // address of collateral token
    address collToken; // "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
    // address of loan token
    address loanToken; // "0xba100000625a3754423978a60c9317c58a424e3D"
    // address of oracle (optional)
    address oracleAddr; // "0x90B9eA01b1c0E2A1b95B03DFBb8E2A4b3F903bfC"
    // min loan amount (in loan token) prevent griefing attacks or
    // amounts lender feels isn't worth unlocking on default
    uint256 minLoan; // "20000000000000000000000"
    // max loan amount (in loan token) if lender wants a cap
    uint256 maxLoan; // "100000000000000000000000"
    // timestamp after which quote automatically invalidates
    uint256 validUntil; // "1735779846"
    // time, in seconds, that loan cannot be exercised
    uint256 earliestRepayTenor; // "0"
    // address of compartment implementation (optional)
    address borrowerCompartmentImplementation; // "0x0000000000000000000000000000000000000000"
    // will invalidate quote after one use
    // if false, will be a standing quote
    bool isSingleUse; // "true"
    // whitelist address (optional)
    address whitelistAddr; // "0x0000000000000000000000000000000000000000"
    // flag indicating whether whitelistAddr refers to a single whitelisted
    // borrower or to a whitelist authority that can whitelist multiple addresses
    bool isWhitelistAddrSingleBorrower; // "false"
}

struct QuoteTuple {
    // loan amount per one unit of collateral if no oracle
    // LTV in terms of the constant BASE (10 ** 18) if using oracle
    uint256 loanPerCollUnitOrLtv; // "800000000000000000"
    // interest rate percentage in BASE (can be negative but greater than -BASE)
    // i.e. -100% < interestRatePct since repay amount of 0 is not allowed
    // also interestRatePctInBase is not annualized
    int256 interestRatePctInBase; // "0"
    // fee percentage,in BASE, which will be paid in upfront in collateral
    uint256 upfrontFeePctInBase; // "40000000000000000"
    // length of the loan in seconds
    uint256 tenor; // "2592000"
}
```

some notes:
- `GeneralQuoteInfo`
  - `minLoan` of 20k $bal may be somewhat high?
  - `validUntil` set to `2025-01-02T01:04:06UTC`?
- ``
  -  `loanPerCollUnitOrLtv` set at 80%
  - `upfrontFeePctInBase` set to 4%
  - `tenor` is 30 days

i understand this prop is a trial, and for that these quote numbers seem fine, but these params should be tuned if we were to increase our deposit

lastly maybe add `0x71E1cc2F7574C798ED893Ef04D55D1E573bE95B1` and `0x1874a08F7975b25944FEb989BbAaA464f61aB3bc` to the address book?

Tritium-VLK:
See:
https://github.com/BalancerMaxis/bal_addresses/pull/257 and https://github.com/BalancerMaxis/bal_addresses/pull/258 for bal_addresses updates.

Pushed a new combined payload that should sim successfully.

gosuto-inzasheru:
it is not simming because of the data structure of the `onChainQuote` parameter. here is a sim anyway, but that fails as well due to too low $bal balance: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/0d153510-f6e7-48e0-bea4-8d9371a8c587

Tritium-VLK (approved):

PR #906, opened 2024-04-21
ci: black workflow

gosuto-inzasheru:


gosuto-inzasheru:
quick lint workflow using black on all python files. commits automatically in case adjustments are needed.

if you agree with this setup, next commit could be a removal of the `src: "./tools/python/helpers/hh_bribs.py"` line in the yaml file, so that the linter will be run on the whole repo

Tritium-VLK (approved):
Looks fine to me, nice.

PR #908, opened 2024-04-22
ci: make black workflow less strict

gosuto-inzasheru:
some brownie related envs still have an older black installed...


Tritium-VLK (approved):

gosuto-inzasheru:
already included in #909

PR #909, opened 2024-04-22
fix: better tenderly sim error handling

### LINKED ISSUE #800: feat: skip tenderly sims if the env vars are not present
this for devs that dont have the prod keys on their local

i made an attempt in #799 but sims were also skipped there in ci 🤔

### LINKED ISSUE #902: bug: Fix tenderly sims
In this RUN ID the reporting failed on tenderly sim:
https://github.com/BalancerMaxis/multisig-ops/actions/runs/8736442395/job/23971305867

In this PR I removed the tenderly sim functionatliy from the reports, and then ran again/everything worked.  I also successfully simmed one of the payloads in tenderly via safe:
https://github.com/BalancerMaxis/multisig-ops/pull/901. - see https://github.com/BalancerMaxis/multisig-ops/pull/901/commits/057b524b28378f859eba9ec982489c1c6d6058c3

Tenderly generation will be globally disabled once that BIP is merged.

gosuto-inzasheru:
closes #902 
closes #800

### DIFF
--- a/.github/workflows/black.yaml
+++ b/.github/workflows/black.yaml
@@ -13,7 +13,7 @@ jobs:
         uses: psf/black@stable
         with:
           options: "--verbose"
-          version: "24.4.0"
+          version: "~= 22.10"
 
       - name: Auto commit
         uses: stefanzweifel/git-auto-commit-action@v5.0.1
--- a/action-scripts/brownie/scripts/get_gauge_mappings.py
+++ b/action-scripts/brownie/scripts/get_gauge_mappings.py
@@ -73,6 +73,8 @@ def gen_report(payload_list):
     reports = []
     for file in payload_list:
         print(f"Processing: {file}")
+        if not file.endswith(".json"):
+            continue
         with open(f"../../{file}", "r") as json_data:
             try:
                 payload = json.load(json_data)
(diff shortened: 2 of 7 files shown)


gosuto-inzasheru:
- graceful catching of either having no tenderly env vars or running into issues with payload json interpretation
- solves the skipping in #903
- the convoluted nested tuple in #904 is not solved, will not spend time on that for now

.github/workflows/black.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
         uses: psf/black@stable
         with:
           options: "--verbose"
-          version: "24.4.0"
```
Tritium-VLK, line 16: Why the change to black version here?  How does this relate to tenderly workflows?

Tritium-VLK (commented):
If you've tested it and it works it looks fine to me.  Maybe run it on a bunch of moved files in 4269-W69?

Why the black version change?

.github/workflows/black.yaml
Tritium-VLK: Figured this out in my next review... #908

Tritium-VLK (approved):

PR #921, opened 2024-04-25
Raise Aura Vote Incentive min to 1500 USDC

Tritium-VLK:
Raise mins such that at least 1500 USD is required to place an incentive on the Aura market.

Authorized as per [BIP-457] with the reason that Aura would like us to try to help them reduce gas costs by avoiding small bribes. 

Simulations: https://docs.google.com/spreadsheets/d/1PDFqNzl_38Wprx38VJ-9vEc_1OUPGwjlzVLZSbm3M2s/edit#gid=2142384993

### DIFF
--- a/config/protocol_fees_constants.json
+++ b/config/protocol_fees_constants.json
@@ -1,6 +1,6 @@
 {
-  "min_aura_incentive": 500,
-  "min_existing_aura_incentive": 500,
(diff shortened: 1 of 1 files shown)


Xeonus (approved):
Raising the min brib amount on the Aura market seems to show good overall results and will improve gas costs for Aura DAO

PR #929, opened 2024-04-27
Github action to built erc20 transfer payloads

Tritium-VLK:
Need to merge before I can test the full action workflow as it doesn't show up until the action json hits main for the first time.


gosuto-inzasheru (approved):
is this much faster than using safe ui to create a payload json..?

PR #938, opened 2024-04-30
BIP-XXX: Enable PAR/EURA gauge incl. gauge cap setting to 2%

Xeonus:
- enable uncapped paUSD/GYD gauge on mainnet
- set relative weight to 2% (after it is enabled)
- Proposal: https://forum.balancer.fi/t/bip-xxx-enable-gauge-for-par-eura-pool-with-a-2-cap-ethereum/5724


gosuto-inzasheru (approved):
`0x10d3eb0400000000000000000000000000000000000000000000000000470de4df820000` decodes to `setRelativeWeightCap(20000000000000000)` which is .02

PR #939, opened 2024-04-30
BIP-XXX: Enable paUSD/stataPolUSDCn gauge incl. gauge cap setting to 2%

Xeonus:
- enable uncapped paUSD/stataPolUSDCn gauge on Polygon POS 
- set relative weight to 2% (after it is enabled)
- Proposal: https://forum.balancer.fi/t/bip-xxx-enable-gauge-for-pausd-statapolusdcn-pool-with-a-2-cap-polygon-pos/5726


gosuto-inzasheru (approved):
rate provider reviewed as safe

PR #954, opened 2024-05-07
Replace signers on new DAO multisigs.

Tritium-VLK:
Using Gnosis-chain as a test safe, also deployed on mode, linea and fraxtal.

Initial setup is 2/8 signers well known to the maxis.  This payload is the template that replaces them with the current dao signers in the usual 6/11 configuration as defined by bal_addresses and on current DAO multisigs. 

Following the test must sign/exec/verify on 3 chains
 - [x] [Fraxtal](https://safe.optimism.io/transactions/queue?safe=fraxtal:0x4f22C2784Cbd2B24a172566491Ee73fee1A63c2e)
 - [x] [Mode](https://safe.optimism.io/transactions/queue?safe=mode:0x4f22C2784Cbd2B24a172566491Ee73fee1A63c2e)
 - [x] [Linea](https://safe.linea.build/transactions/queue?safe=linea:0x4f22C2784Cbd2B24a172566491Ee73fee1A63c2e)


Tritium-VLK:
@mikebmikeb @gosuto-inzasheru mind taking a closer look here to double check?

Test tx on gnosis-chain that we can exec/check once it looks good: https://app.safe.global/transactions/tx?id=multisig_0x4f22C2784Cbd2B24a172566491Ee73fee1A63c2e_0x62a973dfe067956ebb311a0d6c95107595b19aeb2325db8f6a6e39680e190166&safe=gno:0x4f22C2784Cbd2B24a172566491Ee73fee1A63c2e

If you guys can just check the report and/or tenderly and then one of you go ahead and exec, we can check final results.  If that looks good I'll roll out this payload on the other 3 safes for the new chains where we need them.

gosuto-inzasheru (commented):
signers match with the list here: https://docs.balancer.fi/concepts/governance/multisig.html#dao-multisig-signer-set

msig does not exist in address book

Tritium-VLK:
> signers match with the list here: https://docs.balancer.fi/concepts/governance/multisig.html#dao-multisig-signer-set
> 
> msig does not exist in address book

See https://github.com/BalancerMaxis/bal_addresses/pull/291 which is pending everything else being completed/confirmed to merge.

Tritium-VLK:
Please allow @mikebmikeb to do the final merge here after all signers have been changed.

gosuto-inzasheru (approved):
the gnosis msig is not going to be used? might be worth to put it in the address book nonetheless, just for future records

Tritium-VLK:
> the gnosis msig is not going to be used? might be worth to put it in the address book nonetheless, just for future records

Correct on both counts. We may use it later.  I still tthink we should consider using this new safe on all chains for V3 and slowly migrating.

gosuto-inzasheru (approved):

PR #955, opened 2024-05-07
Update GitHub poker

Tritium-VLK:
Include more chains
don't break if something goes wrong
do gas stations too.


Tritium-VLK:
Run: https://github.com/BalancerMaxis/multisig-ops/actions/runs/8988945912/job/24690921279

Tritium-VLK:
Successful bal_addresses generation run with new code. 
 Note that when this is merged the poker should start poking the gas station. 
 https://github.com/BalancerMaxis/bal_addresses/actions/runs/8990601805/job/24696269120

action-scripts/brownie/scripts/pokeInjectors.py
```diff
@@ -3,35 +3,55 @@
...
+        except:
+            print(
+                f"Skipping {chain} because no Address Information can be found in bal_addresses"
+            )
+            continue
+        to_poke = []
+        try:
+            to_poke = book.extras.maxiKeepers.gaugeRewardsInjectors.values()
+        except:
+            print(f"no gaugeRewardsInjectors found in {chain}")
```
gosuto-inzasheru, line 41:
```suggestion
        try:
            to_poke = book.extras.maxiKeepers.gaugeRewardsInjectors.values()
        except:
            to_poke = []
            print(f"no gaugeRewardsInjectors found in {chain}")
```

gosuto-inzasheru (commented):
i dont think `MaxiOps/00NewChainDeployment/replaceDAOTemplateOwners.json` belongs in this pr

action-scripts/brownie/scripts/pokeInjectors.py
gosuto-inzasheru: it is minor, but the idea is that you only need to initialise it if the address book lookup doesnt work

gosuto-inzasheru (commented):
i still think `replaceDAOTemplateOwners.json` should not be in this pr...

Tritium-VLK:
Rebased onto  branch from https://github.com/BalancerMaxis/multisig-ops/pull/958. Merge that before this.

Tritium-VLK:
> i still think `replaceDAOTemplateOwners.json` should not be in this pr...

That got lost in all the noise that shouldn't have been there.  I agree.  Refactored.  Also made it more failsafe.

zkEVM is failing right now due to RPC problems, but they just did a major upgrade yesterday and that's kinda par for the course.  Would be good to get this merged before too long so that when it starts working the injector can fire/rewards can start again.

Ping me when you approve/merge it so I can check on it.

action-scripts/brownie/scripts/pokeInjectors.py
gosuto-inzasheru: way i see it is L40 (`to_poke = book...`) initialises the `to_poke` var. if that action is somehow caught, then `to_poke` needs to be set to `[]`.

your current solution initialises `to_poke` to `[]`, even though the majority of times it will then be reinitialised to `book...`. doesnt make any sense

im suggesting to _try_ to set `to_poke` to `book...`, but if that doesnt work set it to an empty list instead

Tritium-VLK: The way I see it, the script could some day be extended to have additional conditions that could add to the list.  By setting it to empty in any of the conditions you make that less possible.

My general pattern for building lists tends to be initialize it empty and then add to it if there is a reason,a nd if nothing adds to it you have an empty list as a result.

Functionally in this case with the current code it doesn't matter.  If you're stuck on this you're welcome to accept your suggestion and approve the PR.  Otherwise, I still think this makes sense as is and  will probably continue to use this style in the future.  There's been a lot of cases in the past where a choice that was binary took on more options and something zeroing out a list in the middle created bugs/made modifying the code harder.

gosuto-inzasheru: in that case shouldnt you do `to_poke += book.extras.maxiKeepers.gaugeRewardsInjectors.values()` instead? doesnt make any sense either, since it is the initialisation of the var imo

action-scripts/brownie/scripts/pokeInjectors.py
```diff
@@ -3,35 +3,64 @@
...
+        try:
+            book = AddrBook(chain.replace("-main", ""))
+        except:
+            print(
+                f"Skipping {chain} because no Address Information can be found in bal_addresses"
+            )
+            continue
+        to_poke = []
+        try:
+            to_poke = book.extras.maxiKeepers.gaugeRewardsInjectors.values()
```
Tritium-VLK, line 40:
```suggestion
            to_poke += book.extras.maxiKeepers.gaugeRewardsInjectors.values()
```
@gosuto-inzasheru if anything this is the bug  the way I see it.

action-scripts/brownie/scripts/pokeInjectors.py
Tritium-VLK: Pick one and merge it. It's a style thing, the script is simple enough that it can easily be refactored later, you won't miss that.

High level reasoning

If patern is
list = []
maybe add to list
maybe add to list
maybe add to list
maybe add to list

Then I can add a maybe add to list anywhere in there and it works.  The init is clearly different from the adds.

If the pattern is

Maybe add to list or set to 0 if you can't
maybe add to list
maybe add to list
maybe add to list

It's less clear that one shouldn't put a maybe add to list above the add or set.  It's easy if you see a var init line that is simple and know to use that var after.

Tritium-VLK:
@gosuto-inzasheru don't think it was mergable because you were the last commiter and reviewer.  Just merged main to take over last commit.   Also setup automerge.

Finally kicked off this action run to ensure the poker is running.  If it looks good please stamp one more time and done.
https://github.com/BalancerMaxis/multisig-ops/actions/runs/9044611946/job/24853453765

gosuto-inzasheru (approved):

PR #962, opened 2024-05-09
fix: remove oseth from core pools

gosuto-inzasheru:
bip clearly states:

> Note: The owner of the Pool is set to Balancer DAO and the Yield Protocol Fee has been disabled on this pool, thus not making it eligible for Core Pool status.

(ref: https://forum.balancer.fi/t/bip-496-enable-oseth-eth-gauge-ethereum-with-10-emission-cap/5356)


Xeonus (approved):
Removal of core pool entry as per forum post

PR #968, opened 2024-05-14
Reschedule may beets payments to next week to give signers a rest

Tritium-VLK:
There are no other DAO paylaods this week.  Checked with Beets in their discord if this was a problem.   Waiting for confirmation to merge.   May also reschedule next week to W22 when there is already paylaods waiting if there is nothing else this week.


Tritium-VLK:
Beethoven has confirmed this is ok in the optimsm-balancer channel of their discord.

gosuto-inzasheru (approved):

PR #969, opened 2024-05-15
Working Action Payloads: Transfer + add_rewards

Tritium-VLK:
Get the Transfer paylaod action working, and follow the same patter to create an add_rewards action.


Tritium-VLK:
Transfer is working via action.  add_reward is working in dev environment but needs to be merged before the action can trigger/be tested.

gosuto-inzasheru (approved):
approving in order to test in prod

PR #978, opened 2024-05-17
Reschedule KPK and Beets payments for W22

Tritium-VLK:
when there are already payments scheduled.  Just trying to batch.  Shouldn't ask the signers to sign payment only tx's more than once a month at max.

All parties are ok with it.


Xeonus (approved):

Karimsultan (commented):
😒

PR #988, opened 2024-05-25
Bribe_Refund_May2024

Zen-Maxi:
Includes the two txns from May to request 3% refund from Hidden Hand to be sent to mainnet Maxi Multisig: eth:0xc38c5f97B34E175FFd35407fc91a937300E33860


gosuto-inzasheru (approved):
numbers match my query: https://dune.com/queries/3674413

PR #993, opened 2024-05-25
Update 2024-05-25.json

romaleu:
Json


gosuto-inzasheru:
@romaleu what are you trying to do here? afaik `1101` is indeed the correct chain id for polygon zkevm

PR #1002, opened 2024-05-30
BIP-617: Request for BAL to reward UI beta testers

Xeonus:
- Send 1000 BAL from the treasury to the LM multi-sig which will then handle the distribution via Coordinape


gosuto-inzasheru (approved):
https://forum.balancer.fi/t/bip-617-request-for-bal-to-reward-ui-beta-testers/5775

Tritium-VLK (approved):

PR #1017, opened 2024-06-07
non-mergable PR with notes from chat with Jalbrekt

Tritium-VLK:
From our chat.
Mostly about changing directory structure/workflow
Also dump link to trigger relayer in report, and we can work alter to make a comment post from the action with said link.

Translate the YYYY-W## input into YYYY/WW directory


gosuto-inzasheru:
non mergable? what is the purpose of this?

Tritium-VLK:
We had a discussion on a call, I typed some pusedocode/notes inline and then shared this PR into jalbrekts current branch so the notes were clear.

I think the idea is he would reimplement these concepts, this is just here for reference.  It's also not into main branch. 

Can close it whenever, maybe now is good.

gosuto-inzasheru:
will leave it for @jalbrekt85 to close

PR #1051, opened 2024-06-14
fix: chatgpt caffeine deficit

gosuto-inzasheru:
sorry diff looks a bit weird because of linting

only real diff is the outputs parts


Tritium-VLK (approved):

Tritium-VLK:
Thanks.

PR #1053, opened 2024-06-14
chore: vscode linting config

gosuto-inzasheru:
adds vscode linting definitions for python, json and yaml


.github/workflows/build_permissions_bip.yaml
```diff
@@ -4,37 +4,34 @@ on:
...
       PR_NUMBER:
-        description: 'The number of the PR with the a successful Generate Authorizor run'
-        require: true
+        description: "The number of the PR with the a successful Generate Authorizor run"
+        required: true
       BIP_NUMBER:
-        description: 'The number of the BIP to label everything as'
-        require: true
+        description: "The number of the BIP to label everything as"
+        required: true
```
Tritium-VLK, line 11: What is this about, why is the key changing names?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_callinputs

Tritium-VLK (changes requested):

.github/workflows/build_permissions_bip.yaml
Tritium-VLK: I think the change is correct, but it surprises me that a linter is doing it, this is beyond just yaml linting.

gosuto-inzasheru: i changed it, `require` is not correct. my editor threw warnings.

also see your ref; correct keyword is `required`

Tritium-VLK (approved):

PR #1073, opened 2024-06-18
[BIP-634] Funding Balancer OpCo - Admin & Ops - Year 3

Xeonus:
- Fund OpCo for year 3 for 522168 USDC and 12000 BAL
- [Proposal](https://forum.balancer.fi/t/bip-xxx-funding-balancer-opco-admin-ops-year-3/5812)


Zen-Maxi:
I have some confusion here, the proposal mentions Length of Engagement & Budget: 522,168 USDC + 12,000 BAL (12mo). This payload is correct for the USDC only, where should we send the 12,000 BAL? Same address or another? We need to include that to some degree.

Tritium-VLK:
It's not super easy to understand the accounting in this BIP.   In the future, it would be better to lay things out more clearly so a reviewer can check everything adds up internally in the BIP and also that the totals match what is being executed on chain.

I did my best to grok this.  It's good enough, it does what it says on the Tin and the OpCo treasury has demonstrated themselves to be aligned with the DAOs interests and in a sense an extension of the DAO treasury.

Tritium-VLK (approved):

PR #1076, opened 2024-06-19
[BIP-627] Enable syrupUSDC / sDAI Gauge (Ethereum)

Xeonus:
- Enable syrupUSDC/sDAI Composable Stable pool gauge
- [Proposal](https://forum.balancer.fi/t/bip-xxx-enable-syrupusdc-sdai-gauge-ethereum/5815)
- Rate providers checked and verified in registry

PR #1084, opened 2024-06-21
fix: Rate provider reporting on mainnet

Tritium-VLK:
see #1076 for run results from dev station.


gosuto-inzasheru (approved):
nice catch.

`.rstrip("-main")` would have been better, in case there is a `"mainnet-main"` (i dont think there is (yet))

PR #1099, opened 2024-06-26
[BIP-638] Claim and Relock Balancer DAO's Additional AURA/vlAURA

Xeonus:
- relock vlAURA
- claim and lock additional AURA from vesting contract
- both locks will be bundled into the next epoch and will be released as one unlock in 16 weeks.

Note: after this one final claim can be done in 16 weeks, resulting in a total of 2mln AURA tokens locked for the DAO


gosuto-inzasheru (commented):
our unlock takes place at `1720051200` (2024-07-04T00:00:00 UTC)

!!! DO NOT EXEC BEFORE THIS !!!

Tritium-VLK (approved):

Tritium-VLK:
@gosuto-inzasheru pretty sure you can relock the week before your unlock. I'll make sure to sim it and exec when it works.

Issue #1034, opened 2024-06-13
fix: gauge report doesnt retrieve mainnet gauges


gosuto-inzasheru:
eg https://github.com/BalancerMaxis/multisig-ops/pull/1028#issuecomment-2162851020

i think because `mainnet` chain is called `ethereum` in the review registry json

gosuto-inzasheru:
also happens for arb1: https://github.com/BalancerMaxis/multisig-ops/pull/1031#issuecomment-2164561416

gosuto-inzasheru:
solved through #1104

PR #1104, opened 2024-06-26
fix: remove suffix instead of strip

### LINKED ISSUE #1033: fix: gauge report doesnt retrieve rate provider properly on gnosis
see https://github.com/BalancerMaxis/multisig-ops/pull/1027#issuecomment-2162186840

might not be because of the chain but because of being of clp type

gosuto-inzasheru:
closes #1033 #1034 

can be seen working in https://github.com/BalancerMaxis/multisig-ops/pull/1100


Xeonus (approved):
TIL

PR #1107, opened 2024-06-27
[BIP-653] Update-Reward-Distributors-Paladin

Zen-Maxi:
NEEDS AN UPDATE AND INFO FROM PALADIN.
https://forum.balancer.fi/t/bip-xxx-build-a-paylod-for-paladins-bribe-to-liquidity-mining/5835

Need Paladin to validate their desired reward token to take distributorship of on Arbitrum is indeed 0x0945Cae3ae47cb384b2d47BC448Dc6A9dEC21F55. Originally they had 0xcdf7028ceab81fa0c6971208e83fa7872994bee5 but I (Zen) assumed this change on their forum post.

Mainnet payload can interact directly with Authorizer, on Arbitrum i needed to go through AuthorizerAdapterEntrypoint. As for Base i am not sure what they want the gauge the named has no rewards added at this time and desired reward token is the mainnet T address. I expect they just need an add_reward we can do without governance.


gosuto-inzasheru (approved):
matches forum proposal.

i assume `0xfEb352930cA196a80B708CDD5dcb4eCA94805daB` is indeed latest from paladin

PR #1119, opened 2024-07-01
downgrade eth-abi

jalbrekt85:
fixes pip install dependency issue [here](https://github.com/BalancerMaxis/multisig-ops/actions/runs/9636178982/job/26573835683#step:4:91) by downgrading eth-abi to be compatible with safe-eth-py


tools/python/aura_snapshot_voting/vote.py
```diff
@@ -128,14 +128,14 @@ def create_voting_dirs_for_year(base_path, year, week):
     choices = prop["choices"]
     gauge_labels = fetch_json_from_url(GAUGE_MAPPING_URL)
     gauge_labels = {
-        Web3.to_checksum_address(x["address"]): x["label"] for x in gauge_labels
+        Web3.toChecksumAddress(x["address"]): x["label"] for x in gauge_labels
```
gosuto-inzasheru, line 131: it might be better to use `from bal_addresses import to_checksum_address`

ref: https://github.com/BalancerMaxis/bal_addresses/blob/main/bal_addresses/utils.py

gosuto-inzasheru:
note: i also had issues getting this env working (even with this fix) until i downgraded to python 3.10. some issue related to `pysha3`/`safe-pysha3`

jalbrekt85:
> note: i still had issues getting this env working (even with this fix) until i downgraded to python 3.10. some issue related to `pysha3`/`safe-pysha3`

ya safe-eth-py seems to be creating these dependency issues. maybe better to have a separate requirements.txt for `/aura_snapshot_voting`

tools/python/aura_snapshot_voting/vote.py
jalbrekt85: good idea, switched in b377efc

jalbrekt85:
> > note: i still had issues getting this env working (even with this fix) until i downgraded to python 3.10. some issue related to `pysha3`/`safe-pysha3`
> 
> ya safe-eth-py seems to be creating these dependency issues. maybe better to have a separate requirements.txt for `/aura_snapshot_voting`

updated to do this in b377efc

gosuto-inzasheru (approved):
not sure of the added value of the additional reqs.txt. as long as python is on 3.10 all is good with the other reqs.txt as it is in this pr

jalbrekt85:
reverted changes to the original reqs.txt. i think having that separate reqs.txt for voting might avoid dependency issues in the future

gosuto-inzasheru (approved):
gotcha

PR #1123, opened 2024-07-01
fix reporting

Tritium-VLK:


Tritium-VLK:
Need to merge this now to get the payload done. 
This string is somehow becoming a list: https://github.com/BalancerMaxis/multisig-ops/blob/cabf71200feaa23125c7971b6ef944584d8ccc45/BIPs/00batched/2024-W27/1-0x10A19e7eE7d7F8a52822f6817de8ea18204F2e4f.json#L64, it wasn't before... 

Solved the problem by handling if this is a list, but I think TX builder actually needs it to be a string list like this.  I'm kind of confused how this string suddenly becomes a list just now.

Any ideas?

gosuto-inzasheru:
it is due to this [commit](https://github.com/BalancerMaxis/multisig-ops/commit/9454017f45acdcdc813dedcce267bda69252c3b2), apparently `json.load(urlopen(...))` parses the contents a bit better then `json.load(open(...))`

either way solution here is good :)

PR #1131, opened 2024-07-03
[BIP-654] Enable Gauges for baoETH w 2% cap [Ethereum]

Xeonus:
- add baoETH gauges on mainnet
- non YB token, no RP needed
- [Proposal](https://forum.balancer.fi/t/bip-xxx-enable-gauges-for-baoeth-wsteth-baoeth-reth-baousd-susde-baousd-sdai-w-2-cap-ethereum/5848)


gosuto-inzasheru (approved):
capped at 2%, rate providers reviewed safe, tokens match forum prop

PR #1133, opened 2024-07-04
[BIP-655] Enable Initial Fraxtal Gauges

Zen-Maxi:
Proposal: https://forum.balancer.fi/t/bip-xxx-enable-initial-fraxtal-gauges/5859

Rate provider reviews: https://github.com/balancer/code-review/blob/main/rate-providers/FraxtalPriceFeedProvider.md

PR #1135, opened 2024-07-04
feat: add support for new chains

gosuto-inzasheru:
successfully tested on #1133

### DIFF
--- a/action-scripts/brownie/networks.yaml
+++ b/action-scripts/brownie/networks.yaml
@@ -22,3 +22,27 @@ live:
         host: https://rpcapi.fantom.network
         id: fantom-main
         name: Mainnet
+
+  - name: Mode
+    networks:
+      - chainid: 34443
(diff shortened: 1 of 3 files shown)


action-scripts/brownie/networks.yaml
```diff
@@ -22,3 +22,27 @@ live:
         host: https://rpcapi.fantom.network
         id: fantom-main
         name: Mainnet
+
```
Tritium-VLK, line 25: I think if you pull our newest fork branch of brownie these should be here.  Would be good to keep those up to date too. I am not sure anything uses this file, but maybe you set that up at some point.

Tritium-VLK (approved):

action-scripts/brownie/networks.yaml
gosuto-inzasheru: they are not there, but i can pr them later

PR #1136, opened 2024-07-04
Reduce min aura vote to 800 for better handling on L2s and STIP action

Tritium-VLK:
Spoke to @Xeonus about this.  Need min 40k aura for a vote, this should be more than enough.

Right now almost all the aura voting is going ot 1 pool on the L2s, and in some cases there is nowhere for it to go.

Dropping down to 800 should allow a lot more pools to get Aura votes on L2s, and will enable the whole STIP flywheel to work better/spin up smaller pools.

### DIFF
--- a/config/protocol_fees_constants.json
+++ b/config/protocol_fees_constants.json
@@ -1,6 +1,6 @@
 {
   "min_aura_incentive": 1000,
-  "min_existing_aura_incentive": 1500,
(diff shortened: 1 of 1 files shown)


Xeonus (approved):
$500 should result in 43k votes so this threshold should still work given the last voting round results

PR #1204, opened 2024-08-01
[BIP-668]: Revoke Wonderland’s BAL Vest and USDC Payments

Tritium-VLK:
Wonderland agreement has been terminated as of July 31
https://forum.balancer.fi/t/bip-578-engage-wonderland-as-a-service-provider/5659/22?u=tritium


gosuto-inzasheru:
@Xeonus looks like we can actually pass the exact timestamp for the revocation:
```
@external
def revoke(ts: uint256 = block.timestamp, beneficiary: address = msg.sender):
    """
    @notice Disable further flow of tokens and clawback the unvested part to `beneficiary`
            Revoking more than once is futile
    @dev Owner is set to zero address
    @param ts Timestamp of the clawback
    @param beneficiary Recipient of the unvested part
    """
```

gosuto-inzasheru:
aaah never mind it doesnt allow for timestamps in the past

gosuto-inzasheru (approved):

gosuto-inzasheru (approved):

PR #1209, opened 2024-08-05
fix: ignore files removed from pr

gosuto-inzasheru:
reporting should not run for files removed, only added or modified


Xeonus (approved):
great addition. Came up during removal of Wonderland's payloads

PR #1238, opened 2024-08-12
Readme for Final work in LM Multisig for CowAMM launch + send of incentives to Multichain

Tritium-VLK:
- Seed BAL/ETH
- Send remaining incentives budget to omnichain safe also controlled by the maxis. 

The CoW bags are being moved to the omnichain safe to not get mixed up with incomming BAL for an LP operation described by [BIP-669](https://forum.balancer.fi/t/bip-669-provide-80bal-20weth-weth-pol-on-ethereum/5894)


gosuto-inzasheru:
dont understand the need for this since it is lm msig; can be closed imo. nonces have been execd

PR #1246, opened 2024-08-14
[BIP-679] Enable jEUR-EURe gauge [Gnosis]

Zen-Maxi:
Proposal: https://forum.balancer.fi/t/bip-xxx-enable-jeur-eure-gauge-gnosis/5931


gosuto-inzasheru (approved):
why capped?

PR #1262, opened 2024-08-20
ci: make report action fail on error

gosuto-inzasheru:
suggesting to make the report action actually fail on any error, so that merging is prevented in case the report is not successful

payloads should not be approved or merged without the report having run


Tritium-VLK:
In some cases tenderly may not run/etc.  How would we not be blocked by that?  

Would rather test this stuff in a clone repo before rolling out too many changes/restrictions to our current workflow without understanding the full effect.

gosuto-inzasheru:
in which cases would it not run? if it doesnt run, and we havent seen a simulation, why would we be comfortable merging and posting the payload?

Tritium-VLK:
I guess we have tenderly on all chains at the moment now?  Yeah ok.  In the past there have been cases where some chains we operate on just don't have tenderly.  Maybe those days are done.

Tritium-VLK (approved):

PR #1271, opened 2024-08-22
ci: leverage fork var instead of repo names

gosuto-inzasheru:
this should solve the issue of prs opened by actions instead of actual users to be recognised as external prs

### DIFF
--- a/.github/workflows/run_reports_reusable.yaml
+++ b/.github/workflows/run_reports_reusable.yaml
@@ -13,7 +13,7 @@ on:
 jobs:
   generate_reports:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


gosuto-inzasheru:
this was tested in a cloned dummy repo: https://github.com/BalancerMaxis/dummy_repo_ci_test/pulls

external pr from fork indeed did not run the report automatically (due to external environment protection rule), but the pr that resulted from the merge json workflow did run automatically (ran correctly in internal environment)

Tritium-VLK (approved):
I don't really understand how our new CI flow works, but this seems urgent and non-toxic.  Will take time to make sure we all understand soon and try not to stamp to many more of these.

PR #1277, opened 2024-08-23
Change swap fee of 80BAL20WETH-WETH pool from 1% to 0.1%

Xeonus:
- After discussions with Aura contributors, we leave the auraBAL:80BAL20WETH pool as is but only change the swap fee percentage of the 80BAL20WETH:WETH stable pool
- Reduction of swap fees from 1% to 0.1%

This is a first Artifact of swap fee changes in the multi-sig repo. Beneficial for future reference to know what we executed over time.


Tritium-VLK:
0.1% or lower seems reasonable for our POL that is mostly in place to provide utility/UX for those trying to access wrappers.  It may not be the best economic outcome for the DAO, but the purpose of the POL was not to earn money.

Tritium-VLK (approved):

gosuto-inzasheru:
https://etherscan.io/tx/0x3b855c6412e25f6f7b67089c6ac253f4439647fcb469b69de30242ec2427858c#eventlog

PR #1281, opened 2024-08-26
fee change for wrsETH/wETH on OP

franzns:
From 0.4% to 0.2% as an experiment


gosuto-inzasheru (changes requested):
@franzns chain id is set wrong

Tritium-VLK:
@gosuto-inzasheru The validation report used to print if it was available even if the report failed.  This was so you could see things like the chain/etc.  Any idea what happened there?

gosuto-inzasheru:
an actual comment commit only occurs if the error is in the tenderly sim itself.

this payload is just plain wrong since the targeted contract doesnt exist on the specified chain. there is no error handling in place for these kind of mistakes; the action just fails (before https://github.com/BalancerMaxis/multisig-ops/pull/1262 it would do so silently, now it does it noisy)

this could be worth a ticket to explore such error handling, but honest im not sure how much information there is to report on if the report logic is not able to fully run

franzns:
https://app.onchainden.com/safes/oeth:0x09Df1626110803C7b3b07085Ef1E053494155089/transactions/0x45bbfb003cc5373996c0ddf98775a7a155b87c000128eb1e70d9b49050a94124

gosuto-inzasheru (dismissed):

gosuto-inzasheru (approved):
changing from 4bips to 20bips

franzns:
its execd

franzns:
I am unfortunately not authorized to merge :(

But here is the txn link https://optimistic.etherscan.io/tx/0x0810479c87fcfb0dd4e14bf57d0892bf9d11ca458f3d5ebd57da9c81f46ad28e

PR #916, opened 2024-04-25
BIP-607-Update-FTM-Multisig-to-Claim-vested-BEETS

Zen-Maxi:
The DAO Multisig ftm: `0xFe2022da08d17ac8388F5bEFE4F71456255105A4` will call `addOwnerWithThreshold` on itself six times, with the `threshold` equal to `6` and the following new `owner` addresses:

`0x7c2eA10D3e5922ba3bBBafa39Dc0677353D2AF17`
`0xafFC70b81D54F229A5F50ec07e2c76D2AAAD07Ae`
`0xc4591c41e01a7a654B5427f39Bbd1dEe5bD45D1D`
`0x7019Be4E4eB74cA5F61224FeAf687d2b43998516`
`0x11761c7b08287d9489CD84C04DF6852F5C07107b`
`0xcf4fF1e03830D692F52EB094c52A5A6A2181Ab3F`

These can be confirmed via [bal_addresses](https://github.com/BalancerMaxis/bal_addresses/blob/4396ca6ee46272111bb6db65e5054af079e58dda/extras/signers.json#L2)

The DAO Multisig ftm: `0xFe2022da08d17ac8388F5bEFE4F71456255105A4` will then interact with the vesting contract `0x8D381EC09356C3F1805c54B2BB0867e8b417DC3a` calling `release` with the BEETS token address `0xF24Bcf4d1e507740041C9cFd2DddB29585aDCe1e` as the `token` argument.

Once the BEETS is received it will be bridged via Stargate to Optimism to the DAO Multisig on the Optimism network `0x043f9687842771b3dF8852c1E9801DCAeED3f6bc`

PR #1223, opened 2024-08-06
BIP-681-Update-DAO-Multisig-Signers-and-Backup-Signers

Zen-Maxi:
Replaces Trent with Gosuto on all non testnet multisigs. Also transfers 500 BAL to each of 6 designated signer payment addresses.

https://forum.balancer.fi/t/bip-xxx-update-dao-multisig-signers-and-backup-signers/5909

PR #681, opened 2023-12-13
BIP-505: Enable 40DFX-20USDC-40TEL Gauge with 2% emissions cap

Xeonus:
- enable DFX-USDC-TEL gauge with 2% cap on Polygon
- Tenderly-sim: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/0b18793b-a677-4df6-9661-495fa3984fe6

PR #1282, opened 2024-08-26
Removing Fantom from BIP-681

Tritium-VLK:
as it is not a DAO multisig on a active Balancer chain, nor does it have the standard DAO signers as a starting point.
 
 Note that #916 deals with the safe in question, it has still not been executed as the Fantom Safe UI dropped all state in the middle of the signing process and we have not yet taken the time to try again.


Tritium-VLK:
pertains to #681 and

gosuto-inzasheru (approved):
removes fantom msig from [bip681 payload](https://github.com/BalancerMaxis/multisig-ops/pull/1223). fantom is not a prod chain for balancer and its current list of owners is not up to date

PR #1317, opened 2024-09-09
feat: report if gauge is preferential or not

### LINKED ISSUE #1257: Gauge Add Reporting: Check for Preferential Gauge
### Motivation
By improving our gauge add reporting in multi-sig ops, we guarantee that always the correct gauge is set as "preferential" in the subgraph and ultimately in the front-end

### Problem statement
Let's say an entity creates a gauge on Arbitrum, then wants to apply for a BAL gauge and re-creates a new one. In the subgraph and the front-end, the first gauge is automatically set as preferential. One needs to emit an event to set it correctly. Ideally, if this is the case, one should do it before the gauge goes live / is added to the gauge controller so we guarantee frictionless operations

### Implementation details
After checking that root and child-chain gauges are correctly set, do the following:
1. check the root gauge
2. check the getrecipient field
3. on target chain, check subgraph and check preferential gauge field by calling following query:

```
{
  pool(id: "POOL_ADDRESS_LOWERCASE") {
    preferentialGauge {
      id
    }
    gauges {
      id
    }
  }
}
```
4. if getrecipient in root gauge != preferential on target chain, throw error / warning

Details: https://github.com/balancer/gauges-subgraph/issues/91

gosuto-inzasheru:
closes #1257


gosuto-inzasheru:
note the bip683 gauge coming back as not preferential!

action-scripts/brownie/scripts/report_gauges.py
```diff
@@ -425,12 +427,17 @@ def _parse_added_transaction(transaction: dict, **kwargs) -> Optional[dict]:
...
+    )
+
     return {
         "function": f"{to_string}/{command}",
-        "chain": chain.replace("-main", "") if chain else "mainnet",
+        "chain": chain,
         "pool_id_and_address": f"{pool_id} \npool_address: {pool_address}",
         "symbol_and_info": f"{pool_symbol} \nfee: {fee}\na-factor: {a_factor}",
-        "gauge_address_and_info": f"{gauge_address}\nStyle: {style}\ncap: {gauge_cap}",
+        "gauge_address_and_info": f"{gauge_address}\nstyle: {style}\ncap: {gauge_cap}\npreferential: {is_preferential}",
```
Tritium-VLK, line 440: In other cases where there is an error state we are loud about it
!!NOT PREFERENTIAL!!

It's optional, but seems like it would be easy to implement and helps catch bugs.  In a sense could even not show it if there is no problem, but I think we like seeing what is being reported on each time.

gosuto-inzasheru: wdym error? you mean in the case of the gauge not being preferential?

Xeonus (approved):
lgtm, thanks for the report outputs. In this way we will have nice and comprehensive review logs!

PR #1326, opened 2024-09-10
chore: stop poker on gnosis chain

gosuto-inzasheru:
upkeep for our only injector there was configd at https://automation.chain.link/xdai/51275457462711301170724725347664046536233518759579141567815787018122162360515


gosuto-inzasheru:
test run here: https://github.com/BalancerMaxis/multisig-ops/actions/runs/10794561783/job/29939128589

gosuto-inzasheru:
note this will allow us to turn of the zkevm poker in oz defender

action-scripts/brownie/scripts/pokeInjectors.py
```diff
@@ -4,8 +4,7 @@
 
 
 CHAINS_TO_KEEP = [
-    #    "zkevm-main", # currently running in defender due to rpc issues
-    "gnosis-main",
+    "zkevm-main",
```
Tritium-VLK, line 7: The poker is permissioned, so this call will fail now.  

If we want to swap keepers, we need to update the keeper on the underlying contracts and then disable oz + enable zkEVM.

As I recall we're not over our usage limit right now and/or if we are it's not by much. 

Suggest handling zkEVM as a separate ticket/work item and making sure we dot all the Is and cross all the Ts.

Tritium-VLK (changes requested):

action-scripts/brownie/scripts/pokeInjectors.py
gosuto-inzasheru: moved to #1333

Tritium-VLK (approved):

gosuto-inzasheru:
https://app.onchainden.com/safes/gno:0x14969B55a675d13a1700F71A37511bc22D90155a/transactions/0xb1b7f633a4a23282e8e132327c918d90d1bcb88c88234d62caa9c689e9b983c5

PR #1327, opened 2024-09-10
set_reward_distributors_ARB_Gyro_RehypePools

Zen-Maxi:
Goal is to execute on the distributor changes for ARB token. This will allow Gyroscope to stream any ARB they are earning from AAVE distributions of ARB on underlying assets in their pools. Method of execution is going through the ARB multisig to reclaim ownership from the STIP ARB injector, then using the multisig to give ownership to Gyro's provided address. 

List requested by Steffen from Gyro: https://docs.google.com/spreadsheets/d/18y2a_VKkAzVZhxyKE2Ur6VtXI9EVc4Qycr2d1NAQhB0/edit?gid=0#gid=0

PR #1329, opened 2024-09-10
set_reward_distributors_ARB_Gyro_RehypePools

Zen-Maxi:
Goal is to execute on the distributor changes for ARB token. This will allow Gyroscope to stream any ARB they are earning from AAVE distributions of ARB on underlying assets in their pools.

List requested by Steffen from Gyro: https://docs.google.com/spreadsheets/d/18y2a_VKkAzVZhxyKE2Ur6VtXI9EVc4Qycr2d1NAQhB0/edit?gid=0#gid=0


gosuto-inzasheru:
fixed https://github.com/BalancerMaxis/multisig-ops/pull/1327

PR #1331, opened 2024-09-11
Set swap fee to 0.5% for 50BTC.b-50sAVAX

Xeonus:
As per community request set swap fee from 0.1% to 0.5%
image


gosuto-inzasheru (approved):
changes swap fee to 50bips

PR #1338, opened 2024-09-11
REVISED_set_reward_distributors_ARB_Gyro_RehypePools

Zen-Maxi:
Goal is to execute on the distributor changes for ARB token. This will allow Gyroscope to stream any ARB they are earning from AAVE distributions of ARB on underlying assets in their pools. Method of execution is going through the ARB multisig to reclaim ownership from the STIP ARB injector, then using the multisig to give ownership to Gyro's provided address.

List requested by Steffen from Gyro: https://docs.google.com/spreadsheets/d/18y2a_VKkAzVZhxyKE2Ur6VtXI9EVc4Qycr2d1NAQhB0/edit?gid=0#gid=0 - Includes aGHO/GYD Gauge additionally now.


gosuto-inzasheru (approved):
why did you delete the old (already reviewed) payload? it would have been better to just amend that.

here is a diff of the already reviewed payload and the one you are suggesting here: https://www.diffchecker.com/Fvc35FZV/

only changes are the additions of:
- 0xF23d8342881eDECcED51EA694AC21C2B68440929.setDistributorToOwner
- 0x2ffa44fDd19B8d3d2b03842E34F1b4c7E9217591.set_reward_distributor

PR #1362, opened 2024-09-20
Reprogram ARB/USDC injector to add 3k a week to CoWAMM Pendle/ETH for 3 weeks.

Tritium-VLK:


Tritium-VLK:
Request from Paladin/Re.al:
image

Zen-Maxi:
Is this final, it seems we will have an updated schedule from Re.al money team so not going to review until that is clarified and this is updated.

Zen-Maxi (approved):
LGTM 2 rounds for each keeping 5k/wk for ARB and adding 3k/wk for Pendle

PR #1363, opened 2024-09-20
Conclude STIP

Tritium-VLK:
Clear programs from injector and transfer back to omnichain multisig.

PR #1369, opened 2024-09-22
Fix: handle emtpy list in reports

Tritium-VLK:
See #1363 for the issue that caused this fix


Tritium-VLK:
also started a unit_test folder where I am trying to collect 1 of each payload type + whatever odd ones we come across.  This we can copy back and forth to trigger testing, and has better naming/makes more sense than the old test folders.

Tritium-VLK:
cherry picked report from #1369

PR #1376, opened 2024-09-25
[BIP-703] Kill-aUSDC-sDAI-gauge [Gnosis]

Zen-Maxi:
Proposal: https://forum.balancer.fi/t/bip-xxx-kill-ausdc-sdai-gauge-gnosis/6004

The DAO Multisig 0x10A19e7eE7d7F8a52822f6817de8ea18204F2e4f will interact with the AuthorizerAdaptorEntrypoint at 0xf5dECDB1f3d1ee384908Fbe16D2F0348AE43a9eA and call performAction using 0xab8f0945 for the data(bytes) argument and 0xc6710f32c9524b4C6736fEae17BE7c88bB8fF87c as the root gauge address of the aUSDC/sDAI Rehype pool.


gosuto-inzasheru (dismissed):
matches prop

Tritium-VLK:
Next scheduled run is W44 I think. @gosuto-inzasheru try to remember to catch that in review, guess we can also reschedule after the fact. 

Last week of month :)

gosuto-inzasheru (approved):

PR #1392, opened 2024-10-01
Setup first week of rewards for CoWAMM eigen

Tritium-VLK:
5k USD BAL in 1 week


Zen-Maxi (approved):
Matches gauge on frontend, so is confirmed preferential. Report shows distributor and Bal address as correct.

gosuto-inzasheru (approved):

PR #1394, opened 2024-10-01
GHO_USDC_USDT_1000to1250-7days

Zen-Maxi:
Mostly folder re-arranging to have pool parameter changes break down in to amp and swap fee, then by network.

Payload is to update from 1000 -> 1250 in 7 days. GHO is in high demand and therefore relatively low risk as opposed the reverse case which is more typical. Increase requested by ALC (MG).


gosuto-inzasheru (approved):
https://etherscan.io/tx/0xbb90c4e33b7dc1489599f08399f7c7cfb2a33836bbb5c3d3444baf95fad76c83

PR #1407, opened 2024-10-03
special ops: Topup Paladin Direct GHO/USDC/USDT

Tritium-VLK:
**Please confirm GHO is in safe and that tenderly passes and load/sign/exec ASAP.  If you review this and it looks good just load and go**
image


Zen-Maxi (approved):
Perfecto order of ops shows set reward distributor for GHO to LM Multisig, approve GHO and deposit GHO properly to the gauge. Then reset questboard as distributor.

PR #1414, opened 2024-10-09
SWISE_StakewiseEOA-SWISE-osGNO

Zen-Maxi:
Adding SWISE as a reward with Stakewise Safe as distributor for osGNO/SWISE weighted pool on gnosis.

image


gosuto-inzasheru (approved):
whitelists swise safe to distribute $swise to their 20osGNO-80SWISE gauge

PR #1432, opened 2024-10-15
Add APW to Paladin Quest board for APW:WETH 

Xeonus:
Add APW to APW:WETH gauge and enable Paladin Quest board as distributor


gosuto-inzasheru (approved):
apw token is the same as the apw token in the pool

PR #1434, opened 2024-10-15
USDC_Injector-MORE-USDC

Zen-Maxi:
Adding USDC as a reward for the MORE-USDC ECLP gauge on Base. The injector will be Balancer's USDC injector until Paladin QuestBoardV2 is deployed


Zen-Maxi:
Yeah issue is the tools site forks, so I just wont use it any more.

Tritium-VLK (approved):

Issue #1168, opened 2024-07-17
fix: bip report cannot handle arrayed tuple


gosuto-inzasheru:
eg `['scopeFunction(bytes32,address,bytes4,(uint8,uint8,uint8,bytes)[],uint8)']`, see https://github.com/BalancerMaxis/multisig-ops/pull/1154

PR #1443, opened 2024-10-20
Reports: Handle json in payload data.

Tritium-VLK:


gosuto-inzasheru:
related to #1168

Tritium-VLK:
Don't think this is worth doing/too much complexity for very rare use case.

PR #1386, opened 2024-09-27
[BIP-714] Allocate 250,000 USDT to sGYD

JeronimoHoulin:

PR #1444, opened 2024-10-21
fix: move to end of month and put bip# in filenames

gosuto-inzasheru:
https://github.com/BalancerMaxis/multisig-ops/pull/1386 was merged without filenames having bip# and payloads were in the wrong week


Xeonus (approved):
Files in correct execution target folder now

PR #1448, opened 2024-10-21
Add-Franz-to-High-Value-Safes

Zen-Maxi:
Mainnt payroll and Mode LM multisigs adding franz as an additional signer and increasing threshold to 4/7 from 3/6. Additional security measures on safes exposed to over 100k directly due to multiple security issues over the past weeks.

image


Tritium-VLK:
Everything is in bal_addresses.  This was discussed.  LGTM

Tritium-VLK (approved):

PR #1455, opened 2024-10-24
chore: move bip715 to week44

gosuto-inzasheru:
independent audit of the module available here: https://github.com/Recon-Fuzz/aura-locker-module-review


Xeonus:
@gosuto-inzasheru after consulting with BLabs, they clearly stated that they don't want this module to be installed on the DAO multisig. Therefore, executing this BIP next week camunfortunately not happen. 

Before moving any further, we need clear guidelines for the DAO multi-sig and define a strategy for a separate safe holding vlAura and that safe module. Thanks.

PR #1461, opened 2024-10-25
sdBAL-Delegate-to-voting-Optimizer

Zen-Maxi:
Delegating was done via the stakeDAO safe app but I made this payload from scratch to mimic the action and have in repo traceability. The goal is to delegate the sdBAL voting for Maxis in this sdBAL position to passively earn the maximum return on their locked compensation.

This is the interface below seen prior to the safe loading screen. 
image
bal_addresses will be updated accordingly.


Zen-Maxi:
bal_addresses update https://github.com/BalancerMaxis/bal_addresses/pull/501

gosuto-inzasheru (approved):

PR #1468, opened 2024-10-28
fix: shezmu gauges have been killed already

gosuto-inzasheru:
if kept in the w44 folder they become part of the combined payload


Tritium-VLK:
Killed by EmergencyDAO.

Tritium-VLK (approved):

PR #1470, opened 2024-10-28
sUSDE-USDC-100-to-500-7Days

Zen-Maxi:
Per Mike and BD discussion on call to compete for more volume with other sUSDe options on the market. Updates 5x in 1 week. Rule of double per two days is respected.


gosuto-inzasheru (approved):
https://etherscan.io/tx/0x2604b252ee1e2b54de369601d2b493ad574a2a168c731c24f7054e7153ebb40a

PR #1471, opened 2024-10-28
veMODE-BPT-Vote-1

Zen-Maxi:
Voting with 100% of our BPT weight for the Balancer gauge. The gauge is equal to the multisig address in this scenario. 0x9ff471F9f98F42E5151C7855fD1b5aa906b1AF7e

PR #1472, opened 2024-10-28
feat: support simple tuple argument

### LINKED ISSUE #1168: fix: bip report cannot handle arrayed tuple

gosuto-inzasheru:
closes #1168


gosuto-inzasheru:
the simple tuple payload did not sim properly in tenderly before but does now:
- https://github.com/BalancerMaxis/multisig-ops/pull/1471

nested tuples is another issue for later:

Xeonus (changes requested):
For clarification: is the simple_tuple usecase the standard one? Or do we need another "non-tuple" use-case to make sure simple payload reports don't break?

gosuto-inzasheru:
@Xeonus "simple" as in non nested. for nested tuples i opened https://github.com/BalancerMaxis/multisig-ops/issues/1475

Xeonus (approved):
Test case coverage looks good. Works as intended for tuples

PR #1433, opened 2024-10-15
[BIP-726] Adopt SEAL Whitehat Safe Harbor Agreement

DicksonWu654:
- Adopt SEAL Whitehat Safe Harbor Agreement on chain at `0x8f72fcf695523A6FC7DD97EafDd7A083c386b7b6`. Calling `adoptSafeHarbor` with the these parameters [balancerAdoptionParameters.json](https://github.com/user-attachments/files/17383766/balancerAdoptionParameters.json)

PR #1473, opened 2024-10-29
fix: register safe harbor from dao msig

gosuto-inzasheru:
per comment here: https://github.com/BalancerMaxis/multisig-ops/pull/1433#issuecomment-2443114022


Zen-Maxi (approved):
Switching from LM to DAO for safe harbor. Looks good ser.

PR #1482, opened 2024-10-30
[BIP-730] Enable two E-CLP gauges [Ethereum]

Xeonus:
- enable 2 ECLPs on mainnet
- [Proposal](https://forum.balancer.fi/t/bip-730-enable-two-e-clp-gauges-ethereum/6149)


gosuto-inzasheru:
@Xeonus filename does not contain the bip#. not sure if that is an absolute must for non dao msig txs?

gosuto-inzasheru (changes requested):
needs bip# in filename

Xeonus:
@gosuto-inzasheru - this PR was prepared through the Ops UI and I simply forgot to label the BIP-ID, fixing

PR #1483, opened 2024-10-30
auraBAL-Bridge-to-Arbi

Zen-Maxi:
For Q4 2024 to Q1 2025 we will be able to seed the vesters for auraBAL opt in payees. Next step will be to stake auraBAL on Arbitrum, break down amounts proportionally and begin seeding vesters for respective months and time.

Revoke of apporvals of BAL8020 to sdBAL and BAL to Cowswap are included in this bundle for house keeping.

Will not be loaded until approved.


gosuto-inzasheru (approved):
- adding aurabal oft proxy contract to address book here: https://github.com/BalancerMaxis/bal_addresses/pull/509
- `110` is indeed arb1 chain: https://docs.layerzero.network/v1/developers/evm/technical-reference/mainnet/mainnet-addresses
- `decreaseApproval` and `decreaseAllowance` successfully set allowance to `0`
- i would still recommend using the standard `approve(0)` for higher certainty and readability

Issue #1459, opened 2024-10-24
Recycling for Paladin Bribs


Tritium-VLK:
# Context
The omnichain multisig is accumulating bribs from BD voting.  In order for the BD voting to have effect, this value needs to be redirected back into the same pool it was earned from.  BIP-715 specifies that in cases of recycling 30% is taken as a treasury fee and 70% is paid in as bribe markets.  There has been some discussion around a new BIP that proposes 100% is paid in as direct incentives.

# Techincal requirements

## Parser for HH claims API
We need to be able to understand what is available to claim and where it came from so that we can claim it and pay it back in.

Claim data can be found on the Paladin API here: https://api.paladin.vote/quest/v3/copilot/claims/0x9ff471F9f98F42E5151C7855fD1b5aa906b1AF7e

It is a list of json maps shaped like this:
```
 {
      "id": 19198,
      "user": "0x9ff471F9f98F42E5151C7855fD1b5aa906b1AF7e",
      "token": "0x912CE59144191C1204E64559FE8253a0e49E6548",
      "amount": "902503091021552770066",
      "index": 0,
      "period": 1724889600,
      "questId": 11,
      "gauge": "0xB599e35f760c47888aD3fF684FF4cd01ff4Dc45B",
      "proofs": [
        "0xd2f29c757e9a4b26dc8a44d40dfb95cf1583a52e38b2ffb9b00db9dc23e4add5"
      ],
      "path": "bal",
      "distributor": "0xB5757D5D93a26EaA3Bc6b0b25cb2364bE8d5b90E",
      "chainId": 42161
    },
```
    
 We need to 
    1: create a transaction builder  tx that claims everything per chain.
    2: print out a CSV of chain/pool_id/something human readable/amount that shows where redirect needs to go>
    
  ## Redirecting value
  Depending on the method decided, we either need to parse the above CSV, take our 30% cut, and then place bribs on HH , or potentally swap for other coins to avoid overlap and then place schedules on an injector. 
  
Note we should seek to brib on the Balancer market if the bribs are aura bribs to avoid the "double payout" of voters problem.

For now should assume 30% cut and rebrib
  
  ## End result
  
  A workflow that you run, which generates a final payload to claim all available bribs per chain and pay them back out using the desired logic.

Tritium-VLK:
Here is an example claims payload from Paladin:

[...]

PR #1488, opened 2024-11-01
recycle paladin bribes

jalbrekt85:
#1459 - uses SafeTxBuilder to claims paladin bribe and recycle 70% to be deposited on HH for mainnet and arbitrum. 

the script will output two csvs and a payload for each chain: 
- a csv generated from `claim_bribes` that contains relevant info about tokens claimed which is fed into `deposit_bribes`
- a csv generated from `deposit_bribes` that contains the tokens and amounts that should be cowswapped
- a payload with all of the `claim` and `depositBribe` transactions

run:
```sh
python tools/python/paladin_claiming/recycle_bribes.py 
```


tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,196 @@
...
+    chain_addrs = CHAIN_ADDRS[chain_name]
+
+    sell_csv_path = CLAIM_OUTPUT_DIR / f"paladin_tokens_to_sell_{chain_name}.csv"
+    with open(sell_csv_path, "w") as f:
+        f.write("chain,token_address,amount_wei,amount_human,decimals\n")
+
+    bribe_data = {}
+    if chain_name == "arbitrum":
+        # only aura briber on arbitrum
+        aura_briber = SafeContract(chain_addrs["aura_briber"], bribe_market_abi)
```
Tritium-VLK, line 128: This creates this bal conflict overlap thing.  We should ask HH if they can add a bal market on Arbitrum and/or rebrib on Paladin Balancer market.

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,196 @@
...
+        bribe_data["bal"] = (aura_briber, 1)
+        bribe_data["aura"] = (aura_briber, 1)
+    else:
+        bal_briber = SafeContract(chain_addrs["bal_briber"], bribe_market_abi)
+        bribe_data["bal"] = (bal_briber, 2)
+
+        aura_briber = SafeContract(chain_addrs["aura_briber"], bribe_market_abi)
```
Tritium-VLK, line 135: no aura bribs, we're trying to not actually, literally recycle :). If we move them to a new market we will never earn them again.

Tritium-VLK (changes requested):
I didn't test it but code looks good/makes sense.  See notes about bribe markets.  We don't ever want to rebrib on Aura markets in this case.  If we need to we should ask HH to setup a veBAL market, or see if we can figure out how to brib on Paladin with veBAL only.

Can help you get in touch with the right party once we decide a direction.

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: now using hh bal market on arb [0740502](https://github.com/BalancerMaxis/multisig-ops/pull/1488/commits/0740502d06a3cfa22535e1490603f097d1711cd5)

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: fixed in [0740502](https://github.com/BalancerMaxis/multisig-ops/pull/1488/commits/0740502d06a3cfa22535e1490603f097d1711cd5)

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,195 @@
...
+SCRIPT_DIR = Path(__file__).parent
+CLAIM_OUTPUT_DIR = SCRIPT_DIR / "claim_output"
+PAYLOAD_DIR = SCRIPT_DIR / "payload"
+
+CLAIM_OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
+PAYLOAD_DIR.mkdir(parents=True, exist_ok=True)
+
+CURRENT_DATE = datetime.utcnow().strftime("%Y-%m-%d")
+
+w3_by_chain = Web3RpcByChain(os.getenv("DRPC_KEY"))
```
gosuto-inzasheru, line 30: missing `DRPC_KEY=` in `tools/python/.env.template`

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,195 @@
...
+flatbook_mainnet = AddrBook("mainnet").flatbook
+flatbook_arb = AddrBook("arbitrum").flatbook
+
+omni_safe = flatbook_mainnet["multisigs/vote_incentive_recycling"]
+
+CHAIN_ADDRS = {
+    "mainnet": {
+        "paladin_distributor": "0x1F7b4Bf0CD21c1FBC4F1d995BA0608fDfC992aF4",
```
gosuto-inzasheru, line 38: id suggest getting all hardcoded addys in this script to address book (also the ones for arb1)

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,195 @@
...
+        f.write("chain,token_address,amount,amount_mantissa\n")
+
+    # Aggregate claimed bribes by amounts by token and gauge
+    aggregated_bribes = {}
+    with open(csv_path, "r") as f:
+        next(f)
+        for line in f:
+            _, token, gauge, _, amount_mantissa = line.strip().split(",")
+            key = (token, gauge)
+            aggregated_bribes[key] = aggregated_bribes.get(key, 0) + int(
```
gosuto-inzasheru, line 139: not sure if python's `int` has enough precision for some of these mantissas, better to use `Decimal`

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,195 @@
...
+        for token, sell_amount_mantissa in token_sell_amounts.items():
+            token_contract = w3_by_chain[chain_name].eth.contract(
+                address=Web3.to_checksum_address(token), abi=erc20_abi
+            )
+            decimals = token_contract.functions.decimals().call()
+            amount = sell_amount_mantissa / (10**decimals)
+            f.write(f"{chain_name},{token},{amount},{sell_amount_mantissa}\n")
+
+    payload_chain_dir = get_chain_dirs(PAYLOAD_DIR, chain_name)
+    builder.output_payload(
```
gosuto-inzasheru, line 179: can this use json dump with `indent=2` somehow?

gosuto-inzasheru (changes requested):
claiming sim on ethereum: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/40348d56-f50a-4f73-8f71-64e16fe8ab0e

suggesting to at least make the following changes before merging:
- move hardcoded addys to address book
- make sure we have wei precision by implementing `Decimal`

tools/python/paladin_claiming/recycle_bribes.py
gosuto-inzasheru: @jalbrekt85 @Tritium-VLK is maybe also an idea to blacklist these two or three gho and rpl markets to even prevent the need for recycling?

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: added in b77dd9f

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: this would have to be done inside SafeTxBuilder. raised an issue here: https://github.com/BalancerMaxis/bal_tools/issues/54

jalbrekt85:
> ah also getting this in 3.12:
> 
> ```
> multisig-ops/tools/python/paladin_claiming/recycle_bribes.py:28: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
>   CURRENT_DATE = datetime.utcnow().strftime("%Y-%m-%d")
> ```

updated to use datetime.now() 9e33ca7

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: can track here: https://github.com/BalancerMaxis/bal_addresses/pull/513

gosuto-inzasheru: ok great, i left a comment. once merged we can get rid of the hardcoded addys here

tools/python/paladin_claiming/recycle_bribes.py
Tritium-VLK: > @jalbrekt85 @Tritium-VLK is it maybe also an idea to blacklist these two or three gho and rpl markets to even prevent the need for recycling?

Blacklist how?  These are hidden hand bribs.  We are already blacklisted if that is what you mean.

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: f49a48b

tools/python/requirements.txt
```diff
@@ -5,7 +5,8 @@ pandas
 tabulate
 requests
 web3
-git+https://github.com/BalancerMaxis/bal_addresses@0.9.1
+git+https://github.com/BalancerMaxis/bal_addresses@main
+git+https://github.com/BalancerMaxis/bal_tools@main
```
gosuto-inzasheru, line 9: this addition is problematic for my env manager

i think because setup.py in bal_addresses pins bal_tools, and here you are forcing it to main thus creating a conflict

removing the bal_tools line completely fixes it

tools/python/paladin_claiming/recycle_bribes.py
```diff
@@ -0,0 +1,202 @@
...
+    "arbitrum": {
+        "paladin_distributor": flatbook_arb["paladin/DistributorV1"],
+        "bribe_vault": flatbook_arb["hidden_hand2/bribe_vault"],
+        "bal_briber": flatbook_arb["hidden_hand2/bal_briber"],
+    },
+}
+
+
+paladin_distributor_abi = json.load(
+    open("tools/python/abis/MultiMerkleDistributorV2.json")
```
gosuto-inzasheru, line 53: this is confusing. the requirements.txt is `multisig-ops/tools/python/`, so i would consider that to be the root dir and place where the env lives. however, opening it like this implicates we are running from `multisig-ops/` instead?

gosuto-inzasheru (changes requested):

tools/python/paladin_claiming/recycle_bribes.py
Tritium-VLK: That would limit the amount of pools BD can incentivise.  This may make sense, there have been and likely will be technical reasons why it won't always work hence this script.

gosuto-inzasheru:
for ref

Screenshot 2024-11-21 at 16 58 43

https://quest.paladin.vote/#/claim

tools/python/paladin_claiming/recycle_bribes.py
jalbrekt85: i refactored the path handling to work the same way as the aura_snapshot_voting/vote.py script - should make it more clear and will ensure it works when run via gh action - 97108d8

gosuto-inzasheru: works great now!

gosuto-inzasheru (approved):

PR #1491, opened 2024-11-04
chore: claim hedgey op rewards

gosuto-inzasheru:
used ui link shared by zen: https://app.hedgey.finance/claim/f1de8941-e39b-4dc5-af74-4416b2a14150

make sure tenderly is successful since this is only based on raw calldata (and probably not valid anymore once merkl root proof changes)


gosuto-inzasheru:
doesnt work anymore. old sim for ref: https://dashboard.tenderly.co/shared/simulation/1150c4cc-3f54-49cc-a505-9acdd79eba0e

PR #1507, opened 2024-11-22
ci: enable poker on zkevm

### LINKED ISSUE #1333: ci: activate zkevm poker
ref: https://github.com/BalancerMaxis/multisig-ops/pull/1326#discussion_r1752161552

gosuto-inzasheru:
closes #1333 

this will allow us to remove it from defender and thus have all non chainlink keepers in one place

no active schedules on either injector, so no issues with missing a cycle

### DIFF
--- a/MaxiOps/injectorScheduling/zkevm/set_poker_as_keeper.json
+++ b/MaxiOps/injectorScheduling/zkevm/set_poker_as_keeper.json
@@ -1,0 +1,53 @@
+{
+  "version": "1.0",
+  "chainId": "1101",
+  "createdAt": 1732256750075,
+  "meta": {
+    "name": "Transactions Batch",
+    "description": "",
+    "txBuilderVersion": "1.17.1",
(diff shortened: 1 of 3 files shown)


Xeonus:
[Made issue](https://github.com/BalancerMaxis/multisig-ops/issues/1508) for OZ-Automation to make sure it gets decommissioned accordingly after this payload has been executed

Xeonus:
@gosuto-inzasheru I checked our OZ infra and we are poking both injectors [and the gas station](https://github.com/BalancerMaxis/OZ-Automation/blob/0712c1e50fe59e33a615a74e152dc48d947ddfc2/tasks/Polygon%20ZkEVM%20Upkeeper/index.js#L19). I don't see configs for the gas station in this payload. Just want to make sure we are not missing anything. Thanks

Xeonus (approved):
Correct keeper configured. Sim skipped because zkevm is a garbage chain.

PR #1510, opened 2024-11-22
fix: inject fraxscan api key into poker workflow env

gosuto-inzasheru:
poker actions have been failing for the last two days


gosuto-inzasheru:
fixes it: https://github.com/BalancerMaxis/multisig-ops/actions/runs/11970182775/job/33372376375

Xeonus (approved):

PR #1555, opened 2024-12-06
test: deploy an sdbal vester for deployer eoa

gosuto-inzasheru:
i wont add factory to addressbook just yet. deployment log here: https://github.com/gosuto-inzasheru/sd-token-vestor/blob/ef0c75bc9b6a79f11152b34690be22e41a799bb1/broadcast/Deploy.s.sol/1/run-1733473911.json

next step will be to deposit 1sdbal and let it vest over a week or two. both reward claims should be tested in that period


Xeonus (approved):
simple deploy for testing purposes. lgtm

PR #1571, opened 2024-12-10
style: bump black to latest

gosuto-inzasheru:


Xeonus (approved):
version bump and linting. Lgtm

PR #1575, opened 2024-12-11
feat: build merkl airdrop csv based on user pool shares

### LINKED ISSUE #1563: Retroactive Morpho reward Distribution automation
Morpho wants to distribute MORPHO rewards to boosted pool depositors. At the moment, they only distribute rewards every 2 weeks based on the 2 previous weeks deposits into their vaults. Given this, the cleanest solution is to also count boosted pool deposits retroactively and distribute rewards accordingly

Requirements
- Morpho tracks rewards to v3 vault
- Maxis provide code-base to track deposits into boosted pools retroactively (v3 subgraph has pool snapshots)
- Rewards get distributed to depositors either by Morpho or Maxis

Not clear yet:
- If Morpho will send rewards to a new safe or do its own distro based on our airdrop csv
- Angle can't do retroactive campaigns, so we need to do our own airdrop

## Implementation details
Example: deployed test pool on [Gnosis chain](https://test.balancer.fi/pools/gnosis/v3/0xd1d7fa8871d84d0e77020fc28b7cd5718c446522)

Endpoint: https://api.studio.thegraph.com/query/75376/balancer-v3-gnosis/version/latest

Query operation:
```
query Users($where: User_filter, $block: Block_height) {
  users(where: $where, block: $block) {
    shares {
      balance
      id
      user {
        id

      }
    }
  }
}
```

Variables:
```
{
  "where": {
    "shares_": {
      "pool_in": [
        "0xd1d7fa8871d84d0e77020fc28b7cd5718c446522"
      ],
      "balance_gt": 0.0001
    }
  },
  "block": {
    "number_gte": 37435326
  }
}
```

Response:
````
{
  "data": {
    "users": [
      {
        "shares": [
          {
            "balance": "0.210018189788151984",
            "id": "0xd1d7fa8871d84d0e77020fc28b7cd5718c446522-0x75d06bae37a9c349142fe7cee77804900b1c0ec3",
            "user": {
              "id": "0x75d06bae37a9c349142fe7cee77804900b1c0ec3"
            }
          }
        ]
      }
    ]
  }
}
```

gosuto-inzasheru:
closes #1563


tools/python/gen_morpho_airdrop.py
```diff
@@ -0,0 +1,99 @@
...
+    for block in df.columns:
+        # calculate the percentage of the pool each user owns
+        consolidated[block] = df[block] / df[block].sum()
+        # weigh it by the total pool size of that block
+        consolidated[block] *= df.sum()[block]
+    # sum the weighted percentages per user
+    consolidated["total"] = consolidated.sum(axis=1)
+    # divide the weighted percentages by the sum of all weights
+    consolidated["total"] = consolidated["total"] / df.sum().sum()
+    return consolidated
```
gosuto-inzasheru, line 80: @Xeonus or @jalbrekt85 it would be good if one of you has time to draft review this function. it is at the core of this whole feature/pr

if we agree on this piece of logic, then next step for me will be to add multi pool support

Xeonus: Mathematically, looks correct to me.
I see 2 edge cases:
1. might end up with division by zero if `df[block].sum` is 0
2. accumulate of floating point precision issues for small balances but that is a general limiation of data fetching

I would rewrite it to something like this so you don't do the same block sum operation
```
def consolidate_shares(df):
    if df.empty:
        raise ValueError("Empty dataframe provided")
    
    # Pre-calculate sums to avoid redundant computation
    block_sums = df.sum()
    total_sum = block_sums.sum()
    
    if total_sum == 0:
        raise ValueError("No shares found in any block")
    
    consolidated = pd.DataFrame()
    for block in df.columns:
        block_sum = block_sums[block]
        if block_sum == 0:
            continue
        
        # Calculate weighted ownership for this block
        consolidated[block] = (df[block] / block_sum) * block_sum
    
    # Calculate final weighted average
    consolidated["total"] = consolidated.sum(axis=1) / total_sum
    return consolidated
```

gosuto-inzasheru: good point on the zero division, accounted for in 4c29d9c

im not too concerned on the performance optimisations for now; it would save seconds at most (if at all)

also not much to do about floating point precision, given this is how the data is given to start with. but good to keep in mind and to round down at some point to make sure we have enough balance for the final distribution

gosuto-inzasheru:
@Xeonus you mind giving some feedback on current state, maybe also run it locally?

i think the only thing that is really missing atm is epoch management, but it is maybe a bit too early to implement that already

Xeonus:
@gosuto-inzasheru initial thoughts / observations:
1. could run locally successfully. Had to reconfigure web3 dependencies in my venv. might need to check if that is an issue for the action run too
2. Make a more structured config file (similar to watchlist), so it is human readable and we easily see what we create airdrops for:
```
# Current
GAUGES = {
    "0x10A04efbA5B880e169920Fd4348527C64FB29d4D": "0x5bBaeD1fADC08C5fb3e4ae3C8848777E2dA77103",
    "0xc4Ce391d82D164c166dF9c8336DDF84206b2F812": "0x4B891340b51889f438a03DC0e8aAAFB0Bc89e7A6",
}

# Suggestion: 
GAUGE_CONFIG = {
    "pools": {
        "0x10A04...": {
            "gauge": "0x5bB...",
            "description": "Pool name/description"
        }
    }
}
```
3. math checks out for weighted average. How many snapshots do we want to take? One per day might not be enough.

gosuto-inzasheru:
@Xeonus this `GAUGES` dict is a quick fix for the fact that the gauges subgraph is currently not aware of v3 pools. once it is, we can simply retrieve the preferential gauge of the pool without any manual intervention, and thus the gauges dict will become obsolete

re: the amount of snapshot im not sure, it is actually quite fast right now for n=7. given we dont expect that many more pools, we could consider an hourly snapshot or something

gosuto-inzasheru:
@Xeonus this is good to go imo

could rerun report here with proper address labels after https://github.com/BalancerMaxis/bal_addresses/pull/594 has been merged

tools/python/gen_merkl_airdrops.py
```diff
@@ -0,0 +1,229 @@
...
+        params,
+        url="https://api.studio.thegraph.com/query/48427/ethereum-blocks/version/latest",
+    )
+    return int(raw["blocks"][0]["number"])
+
+
+def build_snapshot_df(
+    pool,  # pool address
+    end,  # timestamp of the last snapshot
+    step_size=60 * 60 * 8,  # amount of seconds between snapshots
```
Xeonus, line 100: 8h is on the lower end of cadence. Need to increase to at least 1h in the future when there are more depositors. OK for now

tools/python/gen_merkl_airdrops.py
```diff
@@ -0,0 +1,229 @@
...
+    gauge_shares = {}
+    start = end - EPOCH_DURATION
+    while end > start:
+        block = get_block_from_timestamp(end)
+        pool_shares[block] = get_user_shares_pool(pool=pool, block=block)
+        gauge_shares[block] = get_user_shares_gauge(gauge=gauge, block=block)
+        end -= step_size
+
+    # calculate total shares per user per block
+    total_shares = {}
```
Xeonus, line 115: This implementation currently completely ignores Aura shares and staking! As soon as we have depositors in AURA, we need to add that logic! See [https://app.aura.finance/#/1/pool/245 ](https://app.aura.finance/#/1/pool/245)

tools/python/gen_merkl_airdrops.py
```diff
@@ -0,0 +1,229 @@
...
+            "balance_gt": 0.001,
+            "pool": pool.lower(),
+        },
+        "block": {"number": block},
+    }
+    raw = SUBGRAPH.fetch_graphql_data(
+        "subgraphs-v3",
+        query,
+        params,
+        url="https://api.studio.thegraph.com/query/75376/balancer-v3/version/latest",
```
Xeonus, line 54: Should use decentralized network endpoint for better performance, see [docs](https://docs.balancer.fi/data-and-analytics/data-and-analytics/subgraph.html#v3-subgraphs)

tools/python/gen_merkl_airdrops.py
```diff
@@ -0,0 +1,229 @@
...
+            number
+            timestamp
+        }
+    }"""
+    params = {"where": {"timestamp_lte": ts}}
+    raw = SUBGRAPH.fetch_graphql_data(
+        "blocks",
+        query,
+        params,
+        url="https://api.studio.thegraph.com/query/48427/ethereum-blocks/version/latest",
```
Xeonus, line 92: Will hit rate limits quickly, replace with decentrealized endpoint with ID `9A6bkprqEG2XsZUYJ5B2XXp6ymz9fNcn4tVPxMWDztYC` in the future if we run more snapshots

MaxiOps/merkl/airdrops/morpho-csUSDC-csUSDT-epoch_0.json
```diff
@@ -0,0 +1,23 @@
+{
+  "rewardToken": "0x58D97B57BB95320F9a05dC918Aef65434969c2B2",
+  "rewards": {
```
Xeonus, line 3: If I add up totals I get 1403.29 instead of the supposed 1410.34 as per UI
image

Xeonus (changes requested):

MaxiOps/merkl/airdrops/morpho-csUSDC-csUSDT-epoch_0.json
gosuto-inzasheru: there is a 50bips fee taken on csv airdrops by morpho, see inline comment and or their docs

tools/python/gen_merkl_airdrops.py
gosuto-inzasheru: https://github.com/BalancerMaxis/multisig-ops/issues/1731

tools/python/gen_merkl_airdrops.py
gosuto-inzasheru: i created https://github.com/BalancerMaxis/bal_tools/issues/81

Xeonus (approved):
looks good now. Thanks for all the changes

PR #1580, opened 2024-12-12
Set Reward  Distributors for v3 gauges

Xeonus:
- mainnet: set v1 BAL and USDC injectors as distributors
- gnosis chain: set v2 USDC.e injector as distributor


Xeonus:
had to reconfigure, therefore closing

PR #1449, opened 2024-10-21
Payload for PUR#5 (Mainnet)

JeronimoHoulin:

PR #1581, opened 2024-12-12
Revised Payload for PUR#5 (Mainnet)

dk57:
Maverick stkGHO-GHO pool permissions were removed


Tritium-VLK:
@dk57 should we close the original PR:  #1449

gosuto-inzasheru:
@dk57 the diff with the old payload is huge, not sure why? looks like some ordering changed or something. this forces me to rereview a lot that (probably) has been reviewed already. ref: https://www.diffchecker.com/Yb2IA7yz/

also please move to week 51; this didnt make it to this weekend's voting round

PR #1589, opened 2024-12-16
chore: cancel sablier stream for tritium

gosuto-inzasheru:
cancels per the block it gets execd on, then withdraws all released amount of tokens to the stream recipient

- if this gets executed before eoy, we will need to send the susds delta between cancellation and eoy manually from the misg
- if it gets executed after eoy, tritium will have to send the susds delta between eoy and cancellation manually to the msig

stream id is `1607`; ref: https://app.sablier.com/vesting/stream/LL3-1-1607/


Xeonus (approved):
Correctly calls cancel on stream. Accounting of deltas will be done after execution of this payload.

PR #1597, opened 2024-12-17
[BIP-745] Enable eBTC-WBTC Gauge [Ethereum]

Xeonus:
- enable eBTC-wBTC gaue on mainnet
- [Proposal](https://forum.balancer.fi/t/bip-745-enable-ebtc-wbtc-gauge-ethereum/6233)


gosuto-inzasheru (approved):
note that ebtc does have a rate provider! but it has been reviewed and deemed as safe

PR #1599, opened 2024-12-17
chore: last dao msig exec week is week52

gosuto-inzasheru:


Xeonus (approved):
As discussed, given the low availability from Aura, we are moving payloads back to W52 as there will be no snapshot during the holiday weeks.

PR #1602, opened 2024-12-18
Test Gear Injector on 3pool (mainnet)

Tritium-VLK:


Xeonus:
This is a super old gauge and needs a different config to set rewards ser @Tritium-VLK - I suggest we delete this PR...

PR #1607, opened 2024-12-18
Rewards Injector v1: Reconfigure NET schedule

Figu3:
Cycle 2 of rewards


gosuto-inzasheru (approved):
35k per period, 2 periods

PR #1610, opened 2024-12-19
CowAMM GEAR: Configure GEAR program

Xeonus:
- remove test configuration
- add new recipients and program: 79487.17949 GEAR per week for 2 weeks
- Created new CowAMM [gauge](https://etherscan.io/address/0xA7dc263370Ba6AD86a1971EEcf37e581d5F04E11)

PR #1611, opened 2024-12-19
CowAMM GEAR: Seed Injector

Xeonus:
- Send 158974.37 so we can fund 2 injections, 158974.35 GEAR per injection as per requested program


gosuto-inzasheru (approved):
schedule requires `158974.35898`, this sends a bit of dust to make sure the amount is reached. calcd of of numbers from #1610

gosuto-inzasheru:
superseded by #1614

PR #1623, opened 2024-12-24
style: use vault admin abi on v3 vault

gosuto-inzasheru:
@Zen-Maxi what i did is i used the transaction builder ui to make a call to the vault. however, for the abi field, i pasted the abi from the vault admin as can be found here: https://etherscan.io/address/0x35fFB749B273bEb20F40f35EdeB805012C539864#code

this makes it so that the methods and arguments are still decoded and thus human readable


Zen-Maxi (approved):
Using this format going forward. Merge at will.

Xeonus (approved):

PR #1624, opened 2024-12-24
style: use vault admin abi on v3 vault

gosuto-inzasheru:
same trick as in #1623


Zen-Maxi (approved):
Using this going forward!

Xeonus (approved):

PR #1625, opened 2024-12-26
Boosted-wETH-wstETH-Update1

Zen-Maxi:
A factor to 2500 over 6-7 days to be more competitive than v2 pool (2000). Fee from 2 to 1 bps to be equal to v2 pool.


Xeonus (approved):
Does what it says in description 👍

PR #1626, opened 2024-12-27
GEAR_minWaitPeriod_Update_518400

Zen-Maxi:
To fix injections for cow pool set minWaitPeriod to 7 days or less. Currently too high causing a delay.


Xeonus (approved):
Correctly reduces min wait time between injections

PR #1628, opened 2024-12-29
add-BAL-&-EUL-for-EUL-wETH-CowAMM

Zen-Maxi:
Adds BAL with v1 injector as distributor and EUL with v2 injector as distributor.


Xeonus (approved):
Sets new EUL Injector and adds rewards. EUL injector needs to be added to address book to be decoded correctly

PR #1630, opened 2024-12-30
Continue-USDC-Incentives-2-more-weeks

Zen-Maxi:
2 more weeks of USDC incentives until more pools are created. The plan for this pool was 6 weeks per notion launch strategy, and this original cycle was scheduled for 2 weeks. After 2 more weeks we will access any further need.


Xeonus (approved):
Extension of v3 bootstrapping program. LGTM and executed

Issue #1638, opened 2025-01-03
safe-eth-py doesn't account for queued txs when calculating nonce


jalbrekt85:
safe-eth-py will get the pending nonce for a new tx based on the SafeContract.nonce() call. SafeContract.nonce() is only incremented after a safe tx is executed, so this does not account for other queued txs:
```py
    def retrieve_nonce(
        self, block_identifier: Optional[BlockIdentifier] = "latest"
    ) -> int:
        return (
            self.get_contract()
            .functions.nonce()
            .call(block_identifier=block_identifier)
        )
  ```
this can be fixed by calculating nonce as: `TransactionServiceApi().get_transactions(safe_address)[0]['nonce'] + 1` and passing the result into `build_multisig_tx` instead

PR #1639, opened 2025-01-03
account for pending txs when calculating nonce when posting vlaura vote tx 

jalbrekt85:
#1638


gosuto-inzasheru (approved):
cant test locally rn but lgtm

PR #1642, opened 2025-01-07
[BIP-750] 2025 Gauge Management Proposal

Zen-Maxi:
Three parts to gauge management in Q1 2025.

Define rules going forward for quarterly kills. >60 days without veBAL votes, and <100k USD TVL for >60 days.

Kills initial batch of 102 gauges based on this [table](https://docs.google.com/spreadsheets/d/12yhvFmXKZMwuyb0dUrPB-2WMKiQnHw9L3puDNe6Svjo/edit?usp=sharing). 

Increase weight cap to 100% of ACX/wstETH after incredible core pool round and Binance listing.

Add Balancer v3 Launch Gauges to veBAL system.


gosuto-inzasheru (commented):
- did not review individual adds/kills
- but all adds and kills match gsheet, which has been reviewed independently
- cap increase lgtm

do note that the gauges are currently not seen as preferential, is that a problem @Zen-Maxi? or just a consequence of them being for v3 pools..?

gosuto-inzasheru:
gauges do not come back as preferential because the gauges subgraph is currently unaware of v3 pools. still investigating implications of this

Xeonus:
@gosuto-inzasheru good and valid point in regards to preferential topic

My insights: 
1. my take is data team needs to fix this in the gauges subgraph so they show up correctly as preferential
2. API checks and aggregates subgraph data
3. for checkpointing root gauges: this solely depends on gauge adder and [checkpointer](https://github.com/BalancerMaxis/OZ-Automation/blob/aeb8c23a443c02c3f480ec287ea96081f55f1351/tasks/Mainnet%20Checkpoint%20Root%20Gauges/index.js#L26) contract calls (independent of subgraph) so BAL minting / checkpointing should work
4. only implication this might have is that if we don't fetch the correct preferential gauge, the UI might offer the wrong one for staking

gosuto-inzasheru (approved):

PR #1646, opened 2025-01-08
veMODE-epoch-4

Zen-Maxi:
7500 OP for 3 weeks to Mode uniBTC/mBTC and 7402 OP (remaining balance - hedgey claims) to oBTC/uBTC.

Check sheet [here](https://docs.google.com/spreadsheets/d/1OH98RjvVYzlWlz8aB1JAHB7_d3T4j5kTXRyRmTjCHrk/edit?gid=1036737740#gid=1036737740).


gosuto-inzasheru (approved):
matches numbers from sheet and has same metadata as last epoch

PR #1650, opened 2025-01-10
USDC-Stream-AaveBoostedStablePool-2-weeks

Zen-Maxi:
Continuation for 2 weeks of the same incentive cycle. Hoping to maintain with vlAura in the near future.


Xeonus (approved):
matches previous program

PR #1662, opened 2025-01-14
feat: query pool tokens from v3 vault if v2 fails

### LINKED ISSUE #1572: feat: support gauges for v3 vault
current payload reporting pulls pool information from vault v2, it needs logic to decide between v2 or v3 vault

https://github.com/BalancerMaxis/multisig-ops/blob/c2247dfd3c027f59c29c9b2ac1324683053e6965/action-scripts/brownie/scripts/script_utils.py#L140

gosuto-inzasheru:
closes #1572

i also added a .txt report for bip750 that ran with this change. note that the tokens column is now populated, proving the report is successful in retrieving v3 pool tokens


gosuto-inzasheru:
old report for bip750 is here: https://github.com/BalancerMaxis/multisig-ops/pull/1642

PR #1663, opened 2025-01-14
GHO Buffer initialization

JeronimoHoulin:
To be executed on the managed treasury with the DAO signers.


gosuto-inzasheru (approved):
50k gho is in wallet, sim is now successful

PR #1671, opened 2025-01-15
fix: v3 vault is not deployed on all chains

gosuto-inzasheru:
this should unblock #1668


Xeonus (approved):
Thanks, fixes isse in https://github.com/BalancerMaxis/multisig-ops/pull/1668

PR #1672, opened 2025-01-16
[BIP-762] Update DAO Multisig Signer Set and Finalise 2024

gosuto-inzasheru:
prop: https://forum.balancer.fi/t/bip-762-update-dao-multisig-signer-set-and-finalise-2024/6268


Zen-Maxi (approved):
Correct, getting Lewis from Gyroscope in to each multisig.

PR #1683, opened 2025-01-20
fix: move bip756 into w4 folder with the rest

gosuto-inzasheru:


Xeonus (approved):
Thanks for the fix. Will adhere to this structure with ops ui folder suggestions in the future

PR #1690, opened 2025-01-21
docs: remove issue templates

gosuto-inzasheru:
currently not being used; incentives docs will do a much better job anyway. better to remove this to reduce confusion


Xeonus (approved):
Thanks for the clean-up

PR #1693, opened 2025-01-22
BIP-764-Enable-pxETH-gtWETHe-Gauge-Ethereum

Zen-Maxi:
https://forum.balancer.fi/t/bip-764-enable-pxeth-gtwethe-gauge-ethereum/6307


gosuto-inzasheru (changes requested):
moving 2025 into its own subfolder will break the merge script: https://github.com/BalancerMaxis/multisig-ops/blob/779b4d9c3844440743beddeb0887b4805eff14a4/.github/workflows/merge_json.yaml#L36

maybe better to stick to the bip payload here, and make other structural changes in a different pr?

gosuto-inzasheru:
this moving will also create conflicts with #1692

PR #1695, opened 2025-01-23
feat: get rate providers from v3 pools

### LINKED ISSUE #1670: fix: rate providers for v3 pools are not being retrieved
eg see https://github.com/BalancerMaxis/multisig-ops/pull/1669

gosuto-inzasheru:
closes #1670


Xeonus (approved):
will work for non-nested pools. lgtm

PR #1435, opened 2024-10-15
[BIP-720] Manual Relock of Balancer DAO's vlAura

Xeonus:
- manual relock, needs to be executed after Oct 17th and before Oct 24th 2024

PR #1703, opened 2025-01-27
[BIP-767] Relock Balancer DAOs AURA - Q1 2025

Xeonus:
- BIP ID tbd
- will only correctly sim after timestamp 1738796400 (approx. Thu Feb 6th 2025)
- [Proposal](https://forum.balancer.fi/t/bip-xxx-relock-balancer-daos-aura-q1-2025/6317)


gosuto-inzasheru (approved):
matches payload from [previous manual relock](https://github.com/BalancerMaxis/multisig-ops/pull/1435)

PR #1711, opened 2025-01-28
[BIP-768] Enable xrETH/rETH Gauge on Ethereum

Xeonus:
- enable v3 gauge for xrETH-rETH on mainnet
- [Proposal](https://forum.balancer.fi/t/bip-768-enable-xreth-reth-gauge-on-ethereum/6321)


gosuto-inzasheru (approved):
matches pool from forum prop, rate providers reviewed as safe

PR #1718, opened 2025-01-31
fix: `BIPs/2025/` is for archive purposes only

gosuto-inzasheru:


Xeonus (approved):
Move missing gauge add and gauge removals to week 6. Guarantees DAO multisig bundle execution for kills. Kills are ECLPs with <0.01% voting on veBAL page, so timing is good.

PR #1720, opened 2025-01-31
HiddenHand-Refunds-Dec2024-Jan2025

Zen-Maxi:


gosuto-inzasheru (approved):
all numbers for both december and january match the query: https://dune.com/queries/3674413

PR #1721, opened 2025-02-01
chore: add vote for v3 gho/usdc/usdt

gosuto-inzasheru:


Xeonus (approved):
Nice, correctly replaces v2 with v3 GHO Tripool for updated snapshot

PR #1723, opened 2025-02-03
aPYUSD-apGHO-Amp-to-2000-add-GHO-reward

Zen-Maxi:
Per Tokenlogic's request update A to 2000 over 2 weeks and add GHO as reward for Paladin to distribute.

image


Xeonus (approved):
- a factor change correctly set
- add gho as reward token and Paladins quest board as distributor so Paladin quests can distribute left-over brib rewards to gauge

PR #1725, opened 2025-02-03
chore: was already execd in week5

gosuto-inzasheru:


Xeonus (approved):
correct move of GHO late addition that was executed Friday week 5

PR #1726, opened 2025-02-03
chore: isolate lm payloads for week 6

gosuto-inzasheru:
this way we can wait a bit with dao payloads but at least get lm going


Xeonus (approved):
- correct gauge set
- back-checked gauge adds until 1 month ago. No addition missing

PR #1740, opened 2025-02-05
new protocol fee constants

jalbrekt85:
update to use new fee constants from: https://github.com/BalancerMaxis/protocol_fee_allocator_v2/issues/25


Xeonus:
@jalbrekt85 can you elaborate on this delta? Something that was simply not updated here for a while?

Xeonus:
@jalbrekt85 correct, this can only be merged after we are operational with v3 fee sweeping as per BIP-734.

Makes sense to have a new set of constants for allocator v2 that is also protocol agnostic

jalbrekt85:
@Xeonus alliance config removed. this should be ready to merge

gosuto-inzasheru (approved):

PR #957, opened 2024-05-08
Claim USDC from Cricle CCTP Bridge (manual payload generation)

Tritium-VLK:

PR #1743, opened 2025-02-06
CCTP-Receive-Message-Docs

Zen-Maxi:
Adding in details from Tritium's first pass so everyone can claim CCTP transactions when needed.

https://github.com/BalancerMaxis/multisig-ops/pull/957


Xeonus (approved):
Very nice. Something that we could automate for internal use with Ops UI at some point

PR #1760, opened 2025-02-11
StableSurge-LaunchPools-config

Zen-Maxi:
Initial launch used for testing different pools with 5% or 10% for maxSurgeFee and maxThreshold depending on asset type.

3pools typically should have smaller thresholds due to less room to go off peg. 50% vs 33% for parity require different threshold settings. 5% typically for 3pools. 5-10% being tested for 2pools. 

Max fee settings of 5 and 10% being utilized. Potentially both are too aggressive but we will see if they create to much friction or perform well. Can see potential for lower amounts such as 10 - 100x the baseFee making more sense. This would range from 0.05-4% in most cases as 0.005-0.04 are typical are stable pools.


Xeonus (approved):
- correct setting change payloads
- Maxi omni as "fee setter"

PR #1764, opened 2025-02-12
ci: more compiler cache fixes

gosuto-inzasheru:


Xeonus (approved):
Clear and restore cache configs. Let's see in PROD if they improve performance

PR #1768, opened 2025-02-12
fix: do not use hardcoded pool abis

### LINKED ISSUE #1714: fix: reporting cannot retrieve fee for v3 pools
eg https://github.com/BalancerMaxis/multisig-ops/pull/1712#issuecomment-2619051687

gosuto-inzasheru:
closes #1714

bonus: also fixes an issue with not retrieving rate providers properly


Xeonus (approved):
- generalized abi retrieval
- nested try for v3 fees
- adjustment for rp calls based on generalized abi retrieval

PR #1783, opened 2025-02-17
chore: add payload for sonic dao msig setup

gosuto-inzasheru:
dao msig on sonic still has default signers and low threshold (2/n)

this payload sets signers on par with dao msig on ethereum and raises threshold to 6/n


Zen-Maxi (approved):
LGTM in report.

PR #1755, opened 2025-02-10
[BIP-778] Grant Maxis StableSurge Parameter Change Permissions

Zen-Maxi:
BIP: https://forum.balancer.fi/t/bip-xxx-grant-maxis-stablesurge-parameter-change-permissions/6346

Grants permission of max swap fee and surge threshold on stablesurge pools with swap fee manager delegation to governance.

Perms to cross reference: https://github.com/balancer/balancer-deployments/pull/163/files#diff-17e7d1b01753c19be2eee514c14e09736db3af94d24025c7fd9876c73b6e7950

PR #1784, opened 2025-02-17
[BIP-787] Grant Maxis Amplification Parameter Change Permissions for StableSurge Pools

Zen-Maxi:
…ableSurge-Pools

Using the action-ids from the deployments repo for each network. For example from [Base](https://github.com/balancer/balancer-deployments/blob/bbe1ebb272b8b3085a7b91c33ff997ff72520327/action-ids/base/action-ids.json#L514).


gosuto-inzasheru (commented):
looks good, just needs bip#

couldnt find any other new perms that were introduced by stable surge pools. perms related to the hook were included in bip778 already: https://github.com/BalancerMaxis/multisig-ops/pull/1755

PR #1785, opened 2025-02-17
Set Protocol and Yield Fees: Arbitrum and Base

Xeonus:
- Set global swap fee to 50%
- Set global yield fee to 10%
- Update fee settings on all pools that were created before global fee settings were in effect via ProtocolFeeController call


Zen-Maxi:
Any chance we can leave the USDC/wETH pairs out of this for the updateYieldFee call? The rate provider is toggled to pay yield fees, but if we don’t update it here we will not need to migrate these pools. Is nice for cleaner data collection but not a deal breaker if we view this as "unclean"

Zen-Maxi (changes requested):
Would suggest removing this section from Arbitrum as well, the Arbitrum test pool is subject to the same situation as the base one. 

Line 299:

    {
      "to": "0xa731C23D7c95436Baaae9D52782f966E1ed07cc8",
      "value": "0",
      "data": null,
      "contractMethod": {
        "inputs": [
          {
            "internalType": "address",
            "name": "pool",
            "type": "address"
          }
        ],
        "name": "updateProtocolYieldFeePercentage",
        "payable": false
      },
      "contractInputsValues": {
        "pool": "0x4F073D894d8d3f43810Ff1662D1792D9E32fe2Ec"
      }
    },

Xeonus:
> Would suggest removing this section from Arbitrum as well, the Arbitrum test pool is subject to the same situation as the base one.
> 
> Line 299:
> 
> ```
> {
>   "to": "0xa731C23D7c95436Baaae9D52782f966E1ed07cc8",
>   "value": "0",
>   "data": null,
>   "contractMethod": {
>     "inputs": [
>       {
>         "internalType": "address",
>         "name": "pool",
>         "type": "address"
>       }
>     ],
>     "name": "updateProtocolYieldFeePercentage",
>     "payable": false
>   },
>   "contractInputsValues": {
>     "pool": "0x4F073D894d8d3f43810Ff1662D1792D9E32fe2Ec"
>   }
> },
> ```

@Zen-Maxi  resolved

gosuto-inzasheru (commented):
note that calling `updateProtocolYieldFeePercentage` is permissionless:
```
* @notice Override the protocol yield fee percentage for a specific pool.
* @dev This is a permissionless call, and will set the pool's fee to the current global fee, if it is different
* from the current value, and the fee is not controlled by governance (i.e., has never been overridden).
```

it might be better to explicitly set the protocol fee to `0` on these (test) stablesurge pools?

Zen-Maxi (dismissed):
Looks good, thanks for the help!

gosuto-inzasheru:
rerunning report with updated address book; some pools are currently coming back without a label

gosuto-inzasheru (approved):
we should do a final check once execd if all pools on both arb1 and base were covered

if not, `updateProtocolSwapFeePercentage` should be called from an eoa to sync fee

PR #1791, opened 2025-02-19
Migrate-USDC-V1-injector-to-V2

Zen-Maxi:
Remove GHO from schedule for next 2-3 weeks to align with Aave on larger incentive push.

yUSD strategy raised budget to 500/ 6 weeks total meaning 5 more weeks  from here forward.

Rest are roll over from prior schedule.


MaxiOps/injectorScheduling/arbitrum/Migrate-USDC-V1-injector-to-V2.json
```diff
@@ -0,0 +1,191 @@
...
+        "name": "cancelUpkeep",
+        "payable": false
+      },
+      "contractInputsValues": {
+        "signers": "",
+        "transmitters": "",
+        "f": "",
+        "onchainConfigBytes": "",
+        "offchainConfigVersion": "",
+        "offchainConfig": "",
```
gosuto-inzasheru, line 186: i think all of these args are redundant and can be left out

the method's signature is `cancelUpkeep(id)`

gosuto-inzasheru (commented):
can you share the sheet or doc with the schedules?

gosuto-inzasheru (commented):
- when you say remove gho, you mean the `Aave GHO-USDC-USDT ` pool?
- all periods being one less then in the notion (https://www.notion.so/18de395e22818014b713e5e5f35095a4?v=18de395e228180f7ae12000cf007f5da) is because it injected one round already last week?
- i think we wanted to remove two upkeeps right, the old v1 one and the test v2 one?

Zen-Maxi:
- Yes remove GHO 3pool
- Yes periods are less if rounds fired already, yUSD was extended and only 1 round fired so far.
- Unsure what we "wanted" here my only want was for incentives to reach the LPs on time

gosuto-inzasheru (approved):
- sweep v1 usdc injector
- accept ownership of new v2 usdc injector
- set new schedules on it
- cancel old test upkeep for v2 usdc injector
- cancel deprecated upkeep for v1 usdc injector

PR #1793, opened 2025-02-19
USDQ-USDR-USDC-drop-to-0.01

Zen-Maxi:
Per request from USDR / USDQ representative.

image


gosuto-inzasheru (approved):
set swap fee to .01% (1bip)

PR #1799, opened 2025-02-25
[BIP-789] Grant Maxis Parameter Control on the MEV Tax Hook

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-xxx-grant-maxis-parameter-control-on-the-mev-tax-hook/6367)
- grant MEV tax hook roles on Base (exclusive hook for Base atm) to Omnisig


gosuto-inzasheru (approved):
matches prop

PR #1807, opened 2025-03-04
chore: move some dao payloads to w9

gosuto-inzasheru:


Xeonus (approved):
LGTM, thanks for earlier execution

PR #1810, opened 2025-03-05
test: grant fee processing perms on sepolia

gosuto-inzasheru:


Xeonus (approved):
- Matches sepolia action IDs as per report
- Need withdraw on feesweeper
- order perms on burner

PR #1816, opened 2025-03-10
Gauge-Management-Kills-Q1-2025

Zen-Maxi:
Killing gauges per https://balancer.defilytica.tools/gauge-kill-list which are labeled "To be killed" are included. At risk gauges were removed from the kill list due to not fulfilling the kill criteria.


Xeonus (changes requested):
@Zen-Maxi I think we should also include the wrong duplicate root gauge for the sUSDS:smUSDC pool which is: `0x5670e4e5c8c63c350af2d3cb6b5b1ad19d379947` . See the corresponding vote that introduced it [here](https://forum.balancer.fi/t/bip-783-enable-gauge-for-usdc-susds-boosted-pool-base/6354)

gosuto-inzasheru:
not sure i understand the at risk comment. all gauges in the list with "at risk" status on the dashboard i see in this payload as well

@Xeonus why are they flagged as "at risk" and not "to be killed"? afaics they all adhere to both the >60days since last vote and the maxtvl<$100k constraint

BIPs/2025-W13/BIP-795-Gauge-Management-Kills-Q1-2025.json
```diff
@@ -0,0 +1,858 @@
...
+            "internalType": "bytes"
+          }
+        ],
+        "name": "performAction",
+        "payable": true
+      },
+      "contractInputsValues": {
+        "target": "0x4e15b9184553740f5C822bd3d6169029D302e63e",
+        "data": "0xab8f0945"
+      }
```
gosuto-inzasheru, line 855:
```suggestion
```
these are all redundant, even though @Zen-Maxi writes he did not include them he did

gosuto-inzasheru (changes requested):
almost there

gosuto-inzasheru (approved):

gosuto-inzasheru:
@Xeonus this still needs your stamp as well, given you requested changes

Xeonus (approved):
- Correct list
- added sUSDS duplicate

PR #1817, opened 2025-03-11
[BIP-797] Wire Up V3 Protocol Fee Processing Permissions

gosuto-inzasheru:
https://forum.balancer.fi/t/bip-797-wire-up-v3-protocol-fee-processing-permissions/6411


Xeonus (approved):
- permissions match props in proposal
- correct/identical permission set for each network
- burner functions allow to withdraw to other target if necessary

PR #1834, opened 2025-03-19
[BIP-803] Grant v3 ECLP disable Permissions to Emergency SubDAO wallets

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-803-grant-v3-eclp-disable-permissions-to-the-emergency-dao-multisigs/6427)


gosuto-inzasheru (changes requested):
authorizer on base has its own address!

PR #1837, opened 2025-03-19
chore: add payloads

gosuto-inzasheru:
https://forum.balancer.fi/t/bip-804-protocol-fee-controller-migration/6429

### DIFF
--- a/BIPs/2025-W13/BIP-804-arbitrum.json
+++ b/BIPs/2025-W13/BIP-804-arbitrum.json
@@ -1,0 +1,44 @@
+{
+  "chainId": "42161",
+  "createdAt": 1742379769184,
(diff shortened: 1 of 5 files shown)


gosuto-inzasheru:
@EndymionJkb we werent actually considering sepolia addresses in the address book checks; i pushed a fix

sepolia hasnt been that much a part of our review process, but i think that should change. good step here :)

Xeonus (dismissed):
- Grant admin role to migration contract
- Then finalize migration via omni-multisig
- Action IDs correctly match on all chains

Xeonus (approved):
Reports now correctly pass admin role. Contents didn’t change from last review

PR #1839, opened 2025-03-20
Reconfigure GHO Injector on Base for GHO:USDC

Xeonus:
As per MG request, reconfigure GHO injector on base:
- Increase max injection amount to 50k GHO (was 25k)
- New program: 26k GHO for 1 week to [GHO:USDC](https://balancer.fi/pools/base/v3/0x7AB124EC4029316c2A42F713828ddf2a192B36db) pool
image


gosuto-inzasheru (approved):
increased max injection amount and timestamp `0` should make for a seamless transition to next period

PR #1840, opened 2025-03-20
[BIP-805] Balancer Registry Initialization

gosuto-inzasheru:
to minimise confusion i am still using the week13 folder; suggesting to pr a simple rename of `2025-W13` to `2025-W14` once everything is merged on monday

edit: https://forum.balancer.fi/t/bip-805-balancer-registry-initialization/6430


Xeonus (dismissed):
- Correclty grants admin role to BalancerContractRegistryInitializer

Note that a permissionless call needs to be executed to finalize the initialization as per BIP spec via `initializeBalancerContractRegistry`

Xeonus (approved):
Reports correctly show admin role. Contents didn’t change otherwisw

PR #1854, opened 2025-03-25
ARB USDC Grant: Weeks 3-4 Incentives Program

Xeonus:
- Update USDC ARB grant injector program at 0x08F02E2945C175272Acbb238ACCB6eda56129DE8 
- Arb grant gauges for 1850 each as per [notion](https://www.notion.so/18de395e22818014b713e5e5f35095a4?v=18de395e228180f7ae12000cf007f5da)
- FIX: start timestamp at Thu 5PM CET


Zen-Maxi (dismissed):
Correct 1850 for all pools until we notice the Gyro pairs getting significantly more volume and swap APRs.

Zen-Maxi:
Not really sure the purpose of the timestamp update. Should we not make it asap so the injector just fires when it is next possible on chain? I would rather it set a day in advance than be concerned if there will be a down period.

Xeonus:
@Zen-Maxi yeah in that instance could leave blank. Will fire instantly again as last injection was 17:00:10
For next one can just leave out timestamp but overall not bad practice so one would have time to reconfigure if a bad set slips through. New program live and no downtime, will immediately fire: 

https://balancer.defilytica.tools/rewards-injector/arbitrum/0x08F02E2945C175272Acbb238ACCB6eda56129DE8?version=v2

PR #1860, opened 2025-03-27
Add voting data for 2025-W15

Zen-Maxi:
Add voting data for 2025-W15

PR #1861, opened 2025-03-28
aura voting: add eth-account

jalbrekt85:
the dep changes in #1860 that were needed to get the reporting automation to work broke the posting part. pinning `eth-account` seems to get both working now

also needs to use new [`key`](https://eth-account.readthedocs.io/en/stable/eth_account.signers.html#eth_account.signers.local.LocalAccount.key) attribute


Xeonus (approved):
lgtm

PR #1871, opened 2025-03-31
fix: force creation of cache dirs

gosuto-inzasheru:


Xeonus (approved):
As per feedback of testing routines. Before, script would fail for a novel setup that didn't have cache folders in place

Issue #1396, opened 2024-10-01
style: `payload.json` generated in `/vlaura_voting/.../output` should be linted


gosuto-inzasheru:
else it gets linted in irrelevant prs, eg https://github.com/BalancerMaxis/multisig-ops/pull/1392/commits/46f3b34a7f93ac9e757a2d54a283733497975d42

PR #1875, opened 2025-04-01
prettify json files from post aura guage votes action

jalbrekt85:
#1396 - already using indent=4, this will ensure files are being linted the same way


gosuto-inzasheru (approved):
looking at the files in main branch, it actually looks like they are using indent=2. but yea you are right in that prettier is by default configured to use 4. so not sure where that is being managed, but either way this will solve the issue of linting out of context

PR #1887, opened 2025-04-04
test: only withdraw v3 fees from controller

gosuto-inzasheru:


gosuto-inzasheru:
https://github.com/BalancerMaxis/multisig-ops/actions/runs/14265249309/job/39985514130

gosuto-inzasheru:
not going to work; governance is only granting necessary permissions to the sweeper for the v2 fee controller

PR #1888, opened 2025-04-04
fix: fallback in case no reward data is available

gosuto-inzasheru:


Xeonus (approved):
additional priority fee fallback logic, lgtm

PR #1889, opened 2025-04-04
chore: sweep v1 fee controllers

gosuto-inzasheru:


Xeonus (approved):
We need to manually withdraw fees to the omnisig because we only give perms to fee collector and burner infra for new fee controller v2

PR #1895, opened 2025-04-07
GBPe:sDAI Update Amp factor to 1000

Xeonus:
- As per kpk request, reduce amp factor to 1000
- Ramp down from 4PM CET Apr 7th to Apr 14th (7 days)

image


gosuto-inzasheru (approved):

gosuto-inzasheru:
@Xeonus is there a reason you are doing this one from the lm? in the docs i only see the omni msig having these perms: https://docs.balancer.fi/developer-reference/authorizer/gnosis.html

Xeonus:
@gosuto-inzasheru this is a v2 pool, not a v3 pool. For ComposableStable v6 only LM has perms: https://docs-v2.balancer.fi/reference/authorizer/gnosis.html#last-generated-on-2024-12-04
image

gosuto-inzasheru:
ah ok the reports shows a v3 pool id, that is how i got confused

PR #1897, opened 2025-04-07
sDAI:EURe - Set sDAI Rate Cache Duration to 1

Xeonus:
- As per kpk request set rate cache duration of sDAI to 1 to avoid further leakage because of slow rate updates


Zen-Maxi:
Added in the EURe rate update also. sDAI was all good as it was just need both to have lowest latency possible on our end.

gosuto-inzasheru (approved):
lgtm, looking at the code of `ComposableStablePoolRates.sol` i dont see any issue with setting rate cache duration so low

PR #1898, opened 2025-04-08
Set Rate Cache to 1 for additional Gnosis pools

Xeonus:
GBPe-sDAI: 0x9d93f38b75b376acdfe607cd1ecf4495e047deff
stEUR-EURe: 0x06135a9ae830476d3a941bae9010b63732a055f4

Each token that has a rate provider is set to 1, note that for stEUR-EURe only stEUR has a rate provider configured, so we only set that one

image

The first pool in the request has been set already


gosuto-inzasheru:
@Xeonus does it not also need rate of `1` for the `EURe` component?

Xeonus:
@gosuto-inzasheru the stEUR-EURe pool only has a rate provider for stEUR registered, therefore we only set that one

PR #1927, opened 2025-04-14
GHO-USDC-A-2000-Surge-Params-5%

Zen-Maxi:
Decreasing surge threshold and max fee to 5% per chat with Tokenlogic. Expectation is driving higher swap fees with a higher A factors as well due to routing pools going live using GHO.


gosuto-inzasheru (approved):
- gradually update amp factor to 2000 until `2025-04-19T00:19:50Z`
- update fee% to 5%
- update threshold% to 5%

PR #1929, opened 2025-04-15
GHO-3Pool-3500

Zen-Maxi:
Per MG request.


gosuto-inzasheru (approved):
updating amp factor to 3500 until `2025-04-17T18:56:24Z`

PR #1947, opened 2025-04-23
chore: cancel orders for stuck tokens in burners

gosuto-inzasheru:
orders did not fill due to cow api not being able to quote or incorrect numbers from the subgraph

we will have to process the rescued tokens manually after that on the omni msig


Xeonus (approved):
Successfully executed for fee sweeps

PR #1954, opened 2025-04-24
chore: prep merkl watchlist for epoch 13

gosuto-inzasheru:


Xeonus (approved):
Correctly adds new pool that increased in TVL during epoch 12, will likely have airdrop recievers in epoch 13

PR #1965, opened 2025-04-28
BIP Execution: Contract Registry (BIP-805)

### LINKED ISSUE #1964: BIP Execution: Contract Registry
I've added a "post initialization" section to the document (https://www.notion.so/Balancer-Registry-Initialization-1b9e395e22818008a02dca8288996f7f). We deployed a V2 Stable Pool, which isn't accounted for in the initialization. So - great first test - after the initialization is done, we should add the new V2 pool, and then update the alias to point to it

gosuto-inzasheru:
closes #1964


gosuto-inzasheru:
waiting for feedback from blabs

EndymionJkb:
Looks good to me! This runs the initializer (adding the routers and pool factories). It includes v2 of the standard router and CLR, but because of when this was deployed, v1 of the stable pool. The other calls register v2 of the stable pool, and update the alias for it.

My only question is what do we plan for Avalanche and Optimism? There will be a v2 of the registry initializer soon which includes the v2 stable pool, so no extra calls will be required, that could be run on new chains.

gosuto-inzasheru:
v2 initialiser for avax and opti sounds good!

Xeonus (approved):
Successfully executed on all relevant chains

PR #1971, opened 2025-04-28
docs: accounting for march and april

gosuto-inzasheru:


Zen-Maxi (approved):
Perfect match value wise. New format is good enough.

PR #1979, opened 2025-05-05
alliance config

jalbrekt85:


config/alliance_fee_share.json
```diff
@@ -0,0 +1,58 @@
+{
+  "alliance_members": [
+    {
+      "name": "Rocket Pool",
+      "multisig_address": "0xb867EA3bBC909954d737019FEf5AB25dFDb38CB9",
+      "active": true,
+      "join_date": "2025-04-15",
+      "last_lock_date": "2025-04-30",
+      "pools": [
```
jalbrekt85, line 9: new rocket pools added

Xeonus: I modified the list to be correct as per https://forum.balancer.fi/t/bip-822-rocket-pool-balancer-alliance-program-proposal/6488 - currently we only consider grandfathered in pools, so only the v2 mainnet one qualifies!

Xeonus (approved):

PR #1984, opened 2025-05-06
[BIP-821] Update Permissions to Reflect New Deployment of `ProtocolFeeSweeperV2`

gosuto-inzasheru:


Xeonus (approved):
- matches BIP specs
- correct DAO multi-sigs for all chains
- excludes OP as this is managed by Beets with a different fee processing stack

PR #1892, opened 2025-04-04
[BIP-XXX] Deploy kpk Sub-Roles architecture

JeronimoHoulin:

PR #1989, opened 2025-05-06
Gnosis Chain SubRoles architecture

JeronimoHoulin:
Gnosis Chain payload for Sub Roles Architecture described in: https://forum.balancer.fi/t/bip-xxx-deploy-kpk-sub-roles-architecture/6473


gosuto-inzasheru:
please add to #1892 if it concerns the same bip

JeronimoHoulin:
Added to #1892 !

PR #1992, opened 2025-05-07
[BIP-823] Enable Gauge for Safe Haven QuantAMM BTF

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-823-enable-gauge-for-safe-haven-quantamm-btf/6504)


gosuto-inzasheru (approved):
matches forum prop

PR #2004, opened 2025-05-08
chore: whitelist erc4626 burners

gosuto-inzasheru:


gosuto-inzasheru:
here is a sim of adding the burner and then atomically putting in an order for a erc4626 token: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/e00997fa-5578-46f3-bf65-14e34047b6d1

PR #2005, opened 2025-05-08
[BIP-824] Enable YieldFi yUSD/USDC gauges on Balancer v3 Base

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-824-enable-yieldfi-yusd-usdc-gauges-on-balancer-v3-base-new-pool-deployment/6505)


gosuto-inzasheru (approved):
matches prop

PR #2007, opened 2025-05-08
ci: checkout pr branch not `main`

gosuto-inzasheru:


Xeonus (approved):
make sure main is checked out

PR #2016, opened 2025-05-09
feat: unstuck tokens from both burner types

gosuto-inzasheru:


Xeonus (approved):
- added new burner target for ERC4626
- 5 gwei gas limit

PR #2018, opened 2025-05-09
fix: consider erc4626 asset balance

gosuto-inzasheru:


gosuto-inzasheru:
@Xeonus @jalbrekt85 merging this would be good; some fixes in the bot logic. we should also use these cancelOrder payload that were auto generated; some fees there to unstuck

action-scripts/v3_fee_bot.py
```diff
@@ -100,30 +123,40 @@ def get_pools(chain: str, broadcast: bool = False):
...
             for token in fees["pool"]["tokens"]:
-                erc20 = drpc.eth.contract(
-                    address=to_checksum_address(token["address"]),
-                    abi=json.load(open("action-scripts/abis/ERC20.json")),
-                )
                 for pool_token in pool["poolTokens"]:
                     if pool_token["address"].lower() == token["address"].lower():
                         token_is_erc4626 = pool_token["isErc4626"]
                         break
+                if token_is_erc4626:
```
Xeonus, line 130: what about the issue of "fake" ERC4626 that might cause issues when burning? Can the ERC4626 burner handle those in any circumstance?

Xeonus (approved):
Overall lgtm but need to make sure we don't run into issues with fake ERC4626 tokens

action-scripts/v3_fee_bot.py
gosuto-inzasheru: that is being handled on the burner side:
- https://github.com/balancer/balancer-v3-monorepo/pull/1387
- https://github.com/balancer/balancer-v3-monorepo/pull/1389

PR #2031, opened 2025-05-14
GHO Base Incentive Reschedule

Xeonus:
As per MG request
- 5850 GHO for 1 week for GHO/USDC
- 2150 GHO for 1 week for GHO/USR

Incentives shall fire immediately on each gauge once payload is live


Xeonus:
@zekraken-bot can I get a stamp on this one as it has been already executed? Thanks

PR #2040, opened 2025-05-16
Transfer-50k-OP-Futerarchy-Grant-to-Beets

Zen-Maxi:
Sends 50k OP ear marked for Beets team to fire up growth on their side through June 12 2025.


Xeonus (approved):
Transfer 50k OP to Beets LM

PR #2053, opened 2025-05-21
[BIP-833] USP/aUSDC Gauge proposal

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-xxx-usp-ausdc-gauge-proposal/6538)


gosuto-inzasheru (approved):
addys match screenshot in forum prop

PR #2077, opened 2025-05-28
[BIP-837] Enable wUSDN/BOLD Stable pool Gauge on Balancer v3 Ethereum

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-837-enable-wusdn-bold-stable-pool-gauge-on-balancer-v3-ethereum/6545)


Xeonus:
As per request by the team, the did a last minute retraction. We shall not include this payload in the combined batch next week!

PR #2085, opened 2025-06-02
chore: whitelist mimic burner on avax

gosuto-inzasheru:


Xeonus (approved):
Correctly whitelists mimic burner on fee sweeper v2 on AVAX

PR #2086, opened 2025-06-02
Revert "[BIP-837] Enable wUSDN/BOLD Stable pool Gauge on Balancer v3 Ethereum"

gosuto-inzasheru:
Reverts BalancerMaxis/multisig-ops#2077


Xeonus (approved):
remove omited payload

PR #2088, opened 2025-06-02
chore: set sweeper v2 target token to usdc

gosuto-inzasheru:


Xeonus (approved):
correct method call. Correct [token](https://snowscan.xyz/address/0xb97ef9ef8734c71904d8002f8b6bc66dd9c48a6e) selected

PR #2099, opened 2025-06-05
chore: epoch 18 artifacts

gosuto-inzasheru:


Xeonus (approved):
- Overall Morpho pool TVL and staker count declined, cross-checks match expected amounts
- Lower overall distro amounts
- Correct pool config (no change)

PR #2110, opened 2025-06-10
chore: morpho redist epoch 19

gosuto-inzasheru:


Xeonus (approved):
- pool set correct, overall TVL is rather low on Morpho boosted pools as per June 10th:
image
- base below threshold
- sanity checks for wallets OK

PR #2129, opened 2025-06-17
feat: cancel orders direclty from v3 fee bot eoa

gosuto-inzasheru:
quite a refactor too; wrapping executing tx logic into a func


gosuto-inzasheru:
this proved itself to work flawlessly for last fee run; only issue rn is that in ci cow api gets blocked or something

Issue #2121, opened 2025-06-12
Switch Aura Voter Pipeline to new Locker Safe


Xeonus:
We have successfully locked 2mln Aura in the new locker safe at 0x9a5BDF08a6969A4bDb7724beE3c6d8964BDc0B28

Adjust the aura_snapshot_voting configuration so that it uses the new Aura safe.

https://github.com/BalancerMaxis/multisig-ops/blob/3d4887fb12d36b84c31e793da4dfab9f3efd6120/tools/python/aura_snapshot_voting/vote.py#L44 

--> use `maxi_aura_locker`, see [address_book](https://github.com/BalancerMaxis/bal_addresses/blob/72df2243698e3a6d180689ad8aad3cb89fff9fa6/extras/multisigs.json#L13 )

PR #2130, opened 2025-06-17
replace maxi omni with aura locker

jalbrekt85:
#2121

replaced only in vlaura vote script

there are two other instances of maxi_omni being used:

- here in recycle bribes script (i think this can stay for last run this week?): https://github.com/BalancerMaxis/multisig-ops/blob/a68c919097a6dd7aa7710340e5ad8d38d9b385a3/tools/python/paladin_claiming/recycle_bribes.py#L33
- and in fee bot: https://github.com/BalancerMaxis/multisig-ops/blob/a68c919097a6dd7aa7710340e5ad8d38d9b385a3/action-scripts/v3_fee_bot.py#L20


Xeonus (dismissed):
Ez win

Xeonus (approved):

PR #2147, opened 2025-06-24
[BIP-XXX] Funding for the Balancer OpCo - Admin & Ops - Year 4

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-xxx-funding-for-the-balancer-opco-admin-ops-year-4/6582)
- Pay a total of 372,565 USDC in 4 tranches (quarterly) = 93141.25 per quarter


gosuto-inzasheru:
@Xeonus did you see my comment above?

PR #2166, opened 2025-07-01
feat: blacklist of faulty erc4626 tokens

gosuto-inzasheru:


Xeonus (approved):
Nice addition for smoother fee runs. Known ERC4626 that get stuck are skipped for now.

PR #2191, opened 2025-07-09
Grant 2 Permissions for 0x9e5d6427:  on Optimism

franzns:
This PR manages permissions for 0x9e5d6427:  (0x9e5d64) on Optimism.

Granting 2 permissions.


Xeonus:
@franzns because this is a DAO payload we need to specify if we need to do an execution before EOM, there will be other payloads executed before

@gosuto-inzasheru copilot is a bit disruptive here, reason why reporting is blocked?

gosuto-inzasheru:
@Xeonus the caller address was not checksummed. i enforced the reporting to checksum the address, but probably also a good idea to do so upstream

gosuto-inzasheru:
@franzns is there a reason you also included the old fee controller here? is that still in use on optimism?

franzns:
> @franzns is there a reason you also included the old fee controller here? is that still in use on optimism?

The deployment repo has the same address listed for feecontoller and feecontroller v2: https://github.com/balancer/balancer-deployments/blob/master/addresses/optimism.json because no v1 was deployed.

Guess this is what's "breaking" the reporting here. You can see in the action that only two roles are granted.

gosuto-inzasheru (approved):
ooh i see it was deployed with v2 from the start!

reporting isnt really broken; it is actually doing its work 😄 same action ids correspond to different fee controllers (or at least different labels of fee controllers)

PR #2194, opened 2025-07-09
ci: run fee bot every midnight

gosuto-inzasheru:
requires removal of the review mechanism from the `v3-fee-bot` environment


gosuto-inzasheru:
has its own (protected) branch now

PR #2213, opened 2025-07-15
Fee Bot: Adjustments for triggering

Xeonus:
- adjust fee bot to be manually triggerable but also being able to be called via workflows
- grant permissions to write on the protected branch
- Depends on https://github.com/BalancerMaxis/multisig-ops/pull/2213


gosuto-inzasheru:
why does it need perms to write?

gosuto-inzasheru (approved):
checkout step might still need branch ref but lets see

PR #2214, opened 2025-07-15
Fee Bot: Switch to Scheduling Trigger

Xeonus:
- Exeucte the cron job based on the protected branch setup and workflow
- Inherit secrets from the protected branch
- Needs to be checked-in with https://github.com/BalancerMaxis/multisig-ops/pull/2213


.github/workflows/schedule_v3_fee_bot.yaml
```diff
@@ -0,0 +1,14 @@
+name: Schedule V3 Fee Bot
+
+permissions:
```
gosuto-inzasheru, line 3: what is needed for?

PR #2248, opened 2025-07-24
fix: whitelisting burners on avax should target protocol fee sweeper v1

gosuto-inzasheru:


Xeonus (approved):
Missed this initially, hope with this we can run the fee bot!

PR #2256, opened 2025-07-28
chore: deprecate v3 fee bot

gosuto-inzasheru:
i also removed the secret and its env


Xeonus (approved):
LGTM, secrets also removed

PR #2276, opened 2025-08-05
chore: payload to fwd floating $op

### LINKED ISSUE #2181: Claim DAO OP Rewards for MODE Lock
The Omni-multisig accumulated a lot of OP over the last couple of months without us claiming it. The airdrop campaigns are now handled via Sablier.

Campaign overview: https://app.sablier.com/airdrops/?t=search&c=10&m=0x1fd06c720c18935fd58327a8fef9859445e2d40d&o=true 

Example of Epoch 3 rewards:

Image

We need to check and claim all pending OP rewards as soon as possible

gosuto-inzasheru:
closes #2181


Xeonus (approved):
matches amount and target

PR #2280, opened 2025-08-05
feat: check controller for gauges with same destination

gosuto-inzasheru:


action-scripts/brownie/scripts/reports.py
```diff
@@ -439,18 +495,36 @@ def _parse_added_transaction(transaction: dict, **kwargs) -> Optional[dict]:
     )
     rate_providers_reviews = get_rate_provider_review_summaries(rate_providers, chain)
 
+    # Check for existing gauges with the same destination
+    has_existing_gauge, existing_gauge_addresses = check_existing_gauge_destinations(
```
Xeonus, line 499: There is one scenario where multiple root gauges can exist for one child-chain gauge. As long as there is only one child chain gauge pointing to the pool this is not a race condition and should not trigger a warning as long as there is only one root gauge registered in the gauge controller. One should therefore cross-check the VeBalVotingGauges query for proper validation

action-scripts/brownie/scripts/reports.py
```diff
@@ -73,6 +73,62 @@
...
+    Check if there are existing gauges with the same destination (pool or child gauge).
+    Returns a tuple of (has_existing_gauge, list_of_existing_gauge_addresses)
+    """
+    try:
+        # Query the Balancer v3 API for gauges associated with this pool
+        subgraph = Subgraph(chain)
+
+        # Try to query gauges for this pool using the v3 API
+        query = """
+        query GetPoolGauges($poolId: String!, $chain: GqlChain!) {
```
Xeonus, line 89: One should check if a gauge is already present in the vebalvotinglist and cross-check with the respective pool gauge from the GetPoolGauges endpoint:
- VeBalVotingList query holds all active and killed gauges that have been added or killed via gauge controller:
query VeBalGetVotingGauges {
```
    veBalGetVotingList
    {
        id
        address
        chain
        type
        symbol
        gauge {
            address
            childGaugeAddress
            isKilled
            relativeWeightCap
            addedTimestamp
        }
        tokens {
            address
            logoURI
            symbol
            weight
        }
    }
}
```

This query is much more informative of overall gauge status and also contains context for root and child-chain gauges

Xeonus (changes requested):
Please cross-check if my review comments make sense. I am not fully confident that the current implementation can handle all possible scenarios:

1. multiple root gauges, one child-chain gauge. None of the root gauges has been added to gauge controller yet: pass (this condition is acceptable as long as there will only one root gauge added, this can happen as people created multiple root gauges in the past for the same child by accident)
2. multiple root gauges, one has been killed, the other has not been killed: one shall not add new gauge but unkill the killed one. This data is only present in the veBalVotingList query but not in the pool query present in this code change
3. single root gauge, multiple child-chain gauges, gauge in child-chain set not preferential: fail - this scenario has been covered already

These scenarios are covered in the Ops UI but I don't see them covered in this code base. Especially scenario 1 is OK and should not throw a failure

Xeonus:
@gosuto-inzasheru had to search a bit. If you are looking for an edge case then it's the gauges associated with this pool: https://balancer.fi/pools/base/v3/0x4147d8aeecb75bc9f5c973c3b73d2dfc4dcce131

We had an instance where we had two root gauges for the same pool, one root gauge is killed now which is: 0x5670e4e5c8c63c350af2d3cb6b5b1ad19d379947

PR #2284, opened 2025-08-05
[BIP-865] Grant Emergency subDAO Missing v3 Pool Factory Perms

Xeonus:
- missing pool factory disable permissions for relevant chains
- Proposal as MD file for review as well


gosuto-inzasheru (commented):
grant disable perms for various factories and chains. still needs bip#

how can we do this structurally going forward, so whenever a factory gets deployed? how can we be that including these permissions, the emergency multisig now has all permissions necessary?

gosuto-inzasheru (approved):
https://forum.balancer.fi/t/bip-xxx-grant-emergency-subdao-missing-v3-pool-factory-disable-permissions/6653

PR #2293, opened 2025-08-08
ci: use sha for external branches prs

gosuto-inzasheru:
see https://github.com/BalancerMaxis/multisig-ops/actions/runs/16826878377


Xeonus (approved):
Fixes linting issues from branches

PR #2297, opened 2025-08-10
Update GYD Injector Configuration on base

Lor3mipsvm:
Test: distro of 1 GYD ea to 2 base pools for 1 period


gosuto-inzasheru:
@Lor3mipsvm looks like this pr and #2298 were tests from your side, is it ok to close both?

PR #2300, opened 2025-08-11
chore: final morpho redist epoch28

gosuto-inzasheru:
note that the airdrop for pxeth contained too low values for it to be accepted by merkl

this is also the last morpho redist on eth:, no further rewards available for next epoch (discontinued)


Xeonus (dismissed):
LGTM! Finally done whit this chore!

Xeonus (approved):
lgtm

PR #2330, opened 2025-08-18
chore: merit redists epoch7 ethereum

gosuto-inzasheru:
avax rewards are not available yet


gosuto-inzasheru:
@Xeonus figured it out; some fckry with escaped quotes in the json that python didnt like. safe ui handles it through web3js which doesnt run into issues. anyway fixed

PR #2353, opened 2025-08-25
chore: replace all infura and alchemy references with drpc

### LINKED ISSUE #2308: fix: brownie should not connect through outdated infura rpcs

gosuto-inzasheru:
closes #2308

(repo) secrets should be deleted after merge


Xeonus (approved):
- no more infura refs
- everything replaced with dRPC key, incl. endpoints

Issue #2373, opened 2025-09-01
Upgrade partner config for QuantAMM


Xeonus:
Once the [QuantAMM BIP](https://forum.balancer.fi/t/bip-xxx-balancer-quantamm-protocol-fee-framework/6664) goes live, we need to upgrade the [partner config ](https://github.com/BalancerMaxis/multisig-ops/blob/3875cd2f170fcf4f932993c944c97023dc9676b6/config/alliance_fee_share.json#L86C2-L86C18)to host all QuantAMM pools, see list here:
https://balancer.fi/pools?poolTypes=QUANT_AMM_WEIGHTED

jalbrekt85:
implemented in #2379

PR #2379, opened 2025-09-03
add quantamm pools to partner config

jalbrekt85:
#2373 - adds the two starting quantamm pools. since these are the first pools being added, their fee split will be used as the default

### DIFF
--- a/config/alliance_fee_share.json
+++ b/config/alliance_fee_share.json
@@ -83,7 +83,27 @@
       ]
     }
   ],
-  "partners": [],
+  "partners": [
+    {
+      "name": "QuantAMM",
+      "multisig_address": "0xd785201fd2D9be7602F6682296Bb415530C027Ef",
+      "active": true,
+      "pools": [
+        {
+          "pool_id": "0x6b61d8680c4f9e560c8306807908553f95c749c5",
+          "network": "mainnet",
+          "eligibility_date": "2025-05-07",
+          "active": true
+        },
+        {
+          "pool_id": "0xb4161aea25bd6c5c8590ad50deb4ca752532f05d",
+          "network": "base",
+          "eligibility_date": "2025-05-22",
+          "active": true
+        }
+      ]
+    }
+  ],
   "alliance_fee_allocations": {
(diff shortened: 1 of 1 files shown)


config/alliance_fee_share.json
```diff
@@ -98,16 +118,17 @@
...
+        "vebal_share_pct": 0.0625,
+        "vote_incentive_pct": 0.70,
+        "partner_share_pct": 0.15,
+        "dao_share_pct": 0.0875
+      },
+      "default_non_core": {
+        "vebal_share_pct": 0.25,
+        "vote_incentive_pct": 0.0,
+        "partner_share_pct": 0.50,
+        "dao_share_pct": 0.25
```
jalbrekt85: @Xeonus here is the split that these two pools will use. currently both are being identified as non-core

jalbrekt85:
see test fee run [here](https://github.com/BalancerMaxis/protocol_fee_allocator_v2/pull/207)

config/alliance_fee_share.json
```diff
@@ -99,15 +119,16 @@
...
+      "partner_share_pct": 0.15,
+      "dao_share_pct": 0.0875
     },
     "default_non_core": {
-      "vebal_share_pct": 0.475,
-      "partner_share_pct": 0.35,
-      "dao_share_pct": 0.175
+      "vebal_share_pct": 0.25,
+      "vote_incentive_pct": 0.0,
+      "partner_share_pct": 0.5,
```
Xeonus, line 130: The forum [proposal](https://forum.balancer.fi/t/bip-xxx-balancer-quantamm-protocol-fee-framework/6664) states 80% fee share to QuantAMM for non-core pools, so this should be adjusted accordingly.

I think it should be:
partner_share_pct: 0.8
vebal_share_pct: 0.165 (0.2 * 0.825)
dao_share_pct: 0.035 (0.2 * 0.175)

Xeonus (changes requested):
Please review the non-core config and adjust accordingly based on BIP specs

PR #2412, opened 2025-09-16
restructure partner fee configuration for dynamic pool discovery

jalbrekt85:
changes needed to fee config in order to implement https://github.com/BalancerMaxis/bal_tools/issues/128 into fee allocator

separates the alliance and partner config into their own files. partner pool list is replaced by a `pool_types` field. the allocator will automatically fetch partner pools based on configured pool type for each partner


Xeonus (approved):
- removes static core pool alloc in file and fetches dynamically in fee allocator
- dedicated partner config resulting in easier segregation of concerns

LGTM!

PR #2416, opened 2025-09-18
fix: safe ui does not accept structs

gosuto-inzasheru:


gosuto-inzasheru:
still some json parsing fckry

PR #2447, opened 2025-10-01
Combined payload: remove OpCo DAO payments

Xeonus:
Moving OpCo payment by 1 week as DAO EOM bundle will execute/be loaded in W41

EDIT: as agreed with the Foundation, this payload was overwritten by BIP-873. Therefore remove its payload from the weekly batch


Xeonus:
After discussion with Danko, we need to cancel this payment. It was overwritten with [BIP-873](https://forum.balancer.fi/t/bip-873-balancer-ecosystem-roadmap-proposal-and-funding/6668). Meaning we need to remove this payload entirely. Any left-over USDC in the DAO treasury shall be sent to the managed multi-sig and OpCo will request funds from there

PR #2459, opened 2025-10-07
fix: ezkl poolset creation is dependent on dao exec

gosuto-inzasheru:
https://github.com/BalancerMaxis/multisig-ops/pull/2458#issuecomment-3376338979


Xeonus (approved):
Correctly moves EZKL config payload for Omni out. Shouldn’t have been part of this week‘s omni batch in the first place!

PR #2462, opened 2025-10-08
feat: allow bot to sweep on plasma

gosuto-inzasheru:


Xeonus (approved):
Correct role to sweep directly without a burner in place

PR #2470, opened 2025-10-09
chore: merit redists avax round 22

gosuto-inzasheru:


Xeonus (approved):
Already executed, so couldn't verify amounts as they were distributed

Issue #2443, opened 2025-09-29
chore: redist Merkl rewards on Plasma


Xeonus:
For the first week, Plasma incentives were redistributed to our Omni-sig. Issue was same as with Base where the same underlying market overlaps with different pools (for USDT).

We agreed with Merkl to do a one time retro-active redistribution for the first week of incentives until Thursday Oct 2nd. Afterwards, Merkl has fixed their code base to correctly forward rewards directly to users.

PR #2477, opened 2025-10-14
feat: redist merkl rewards plasma

gosuto-inzasheru:
#2443

this is just a redirect of the rewards of the first two weeks; claim and immediately create a campaign that will redist based on merkl's internal fwding rules

rewards from the third week onwards are already being processed in this way


Xeonus:
> create a campaign that will redist based on merkl's internal fwding rules

Can't review without knowing these internal rules. Couldn't find hints in Merkl chat

Xeonus (approved):
Inactive campaign view actually states 9th of October, so time-span is correct:
image

PR #2487, opened 2025-10-22
chore: whitelist sweeper v2 on plasma

gosuto-inzasheru:
😑


Xeonus (approved):
Adding missing perm so we can actually process fees via the feesweeper-v2

PR #2497, opened 2025-10-28
chore: plasma redist - update proofs, timestamps and amount

gosuto-inzasheru:
again; might need final update of proofs right before signing/execing


Xeonus (approved):
LGTM

PR #2512, opened 2025-11-03
Emergency payload kill Balancer v2 stable gauges

Xeonus:
- kill all currently active stable pool gauges still active for Balancer v2
- Apply filter to only kill stable pools that have rate providers (or have been paused)
- keep weighted pools as is
- exclude any veBAL wrapper that might still be active (e.g. tetu, sdBAL etc)

Emergency subDAO has authorization as described in [docs](https://docs.balancer.fi/concepts/governance/emergency.html#specifications)


Xeonus:
@zekraken-bot good point, I need to remove gauges that haven't been added to the veBAL voting list for consistency. Those are just stale and will never be voted in anyways.

PR #2516, opened 2025-11-04
Emergency action: Disable CSPv6 pool factories

Xeonus:
- Given the recent exploit is targeting CSPv5 and CSPv6, disable CSPv6 pool factories
- CSPv5 has already been disabled after introduction of v6


gosuto-inzasheru (approved):
indeed bnb and sepolia missing which is fine imo---all other chains are accounted for

PR #2526, opened 2025-11-10
[BIP-882] Safe Signer Swaps

Xeonus:
As per BIP-882 we will implement the nested safe setup with Balancer Onchain Ltd and the new operator safe as signers of all operational safes. Therefore, the signer swap will:
- reduce threshold to 1/2 on all those safes
- set Balancer Onchain Ltd and Operator safes as signers

Safes that need the swap:
- Omnisig
- Aura locker
- fees multisig
- gauntlet safe onn mainnet
- LM multisig


gosuto-inzasheru:
hyperevm is chain id `999`

gosuto-inzasheru (dismissed):
all payloads add maxyz operator and balancer onchain ltd safes as a 1/2 signer, then remove the rest of the signers. some slight differences per safe/chain:

### aura_locker
- no deviations

### fees
- no deviations

### lm
- gnosis has one extra signer that needs to be removed ✅

### omni
- mode has one extra signer that needs to be removed ✅

note: all payloads make an additional unnecessary call to `changeThreshold` (already taken care of with the `_threshold` arg in `removeOwner` call); could save some gas there

Xeonus:
All signer swaps executed except MODE. Performing token transfers / unstake of vePositions first

PR #2527, opened 2025-11-10
chore: fix file fetch logic circumvent rate limits

Xeonus:
- Should fix issues with this PR https://github.com/BalancerMaxis/multisig-ops/pull/2526


Xeonus:
#2528 is the more elegant solution. Therefore, this is not needed anymore

PR #2530, opened 2025-11-12
perf: move disable calls to emergency msig where possible

gosuto-inzasheru:
ethereum is also possible from blabs_ops instead of dao, which would alleviate dao signer load and would have my preference. not sure about that msig though


Xeonus (approved):
On some L2's Emergency has ability to pause the LBP factory, reduces load on DAO multi-sig

PR #2535, opened 2025-11-17
X-Layer Nested Safe setup: Signer Swap

Xeonus:
[BIP-879](https://forum.balancer.fi/t/bip-879-deploy-balancer-v3-on-x-layer-okx-chain/6843) voted in Balancer v3 deployment on X-Layer. All safes have been deployed. This payload will:
- Perform nested safe setup / signer swap with Operator and Balancer Onchain Ltd safes as signers
- adhere to BIP-882 signer swap / signer setup


Xeonus:
[Executed](https://www.oklink.com/x-layer/tx/0x5de60c8408be5f583c0ea968e575a5932e8326846b564e52a92d30dbb156e102)

PR #2552, opened 2025-11-28
Balancer v2 Factory disables: move LBPs factory disable on Gnosis

Xeonus:
Circles infrastructure relies on NoProtocolFeeLiquidityBootstrappingPoolFactory to function. They will migrate to v3 but this takes another 2-3 months. We shall not shut this factory down until they migrated to the new infrastructure!

image

Moving disable call to end of Q1 2026


Xeonus:
As for now timelines are unclear I would rather not include it in any week and execute by mistake

PR #2576, opened 2025-12-11
chore: replace locker module v2 with v3

raq0x:


gosuto-inzasheru (commented):
disable module v2, enable module v3 (correct address)

should we also add the necessary payload here for the chainlink automation?

PR #2597, opened 2025-12-24
Q2 BAL payment to kpk treasury

taravellokpk:
This PR adds a one-off operational payload to execute the final Q2 BAL payment to the kpk treasury.


gosuto-inzasheru:
looks good @taravellokpk! can you move it to `TreasuryOps/` though? we have created that folder for the new treasury signer setup

PR #2608, opened 2026-01-07
fix: generate base redist using new airdrop ui 

gosuto-inzasheru:


Xeonus (dismissed):
- new amounts after re-claim confirmed
- sanity checked the 3 biggest airdrop campaign creation strings and could confirm correct payload layout with new creation UI and payload layout (uses EngineCampaign field etc) [here](https://studio.merkl.xyz/create-campaign)
Example: tuples match (except creation timestamp):

Image

gosuto-inzasheru:
execd

PR #2610, opened 2026-01-07
[BIP-900] Enable Gauge for 60GOLD/40wstETH Pool at 10% cap

Xeonus:
- [Proposal](https://forum.balancer.fi/t/bip-xxx-enable-gauge-for-60gold-40wsteth-pool-at-10-cap/6912)


gosuto-inzasheru (approved):
matches prop

PR #2640, opened 2026-01-21
Monad: Omni signer swap

Xeonus:
- All safes initialized
- Replace previous signers with operator and balancer onchain ltd safes as with other chains

depends on https://github.com/BalancerMaxis/bal_addresses/pull/856

brownie dependencies for MONAD support not updated yet!


gosuto-inzasheru (approved):
lgtm

PR #2629, opened 2026-01-15
Merkl Wrapper Token Distribution Test

Xeonus:
Create a test campaign for rETH distribution via the omni on mainnet
- Campaign setup description found [here](https://docs.merkl.xyz/merkl-mechanisms/features#how-it-works)
- Omni is pre-seeded with 0.03 rETH
- Test wallets able to claim are zekraken deployer, xeonus deployer and juanjo's EOA

Files added:
- airdrop file distributing 0.0005 rETH to 3 wallets
- combined payload with necessary approvals and campaign creation (via https://studio.merkl.xyz/create-campaign/airdrop)
image

rETH test wrapper: `0x289c74E8E83B95C2B6e166E02cD8a90dd52d492A`

PR #2641, opened 2026-01-21
Merkl Wrapper Token: wstETH Test Campaign

Xeonus:
- Exact same setup as with rETH airdrop (https://github.com/BalancerMaxis/multisig-ops/pull/2629)
- wstETH wrapper token: 0xc63FF8aF78414F90069aF3558D939127625ABE78

image


gosuto-inzasheru (approved):
wrapper and sim lgtm, but fee calc should have been fixed by now. maybe ping angle?

PR #2656, opened 2026-01-29
feat: mint osgno

gosuto-inzasheru:


gosuto-inzasheru:
already performed by kpk via roles module

PR #2664, opened 2026-02-03
[BIP-882] Update Treasury Signers on Polygon

### LINKED ISSUE #2663: [BIP-882] Update Treasury Signers on Polygon
even though the managed treasury safe on polygon was technically never in use, it was deployed (`matic:0x0EFcCBb9E2C09Ea29551879bd9Da32362b32fc89`). for future compatibility, it should be switched to the treasury council set as defined in bip882, similar to the ones that are already in use (`eth`, `gno` and `arb1`)

- generate payload that switches signers (see `BIPs/2025-W45/BIP-882-Onchain-Migration` for payloads other chains)
- add to address book accordingly

gosuto-inzasheru:
closes #2663


raq0x (approved):
Same as in original BIP-882 payloads. LGTM 👍

PR #2673, opened 2026-02-10
chore: reschedule dao payloads to eom and treasury to this week

gosuto-inzasheru:


gosuto-inzasheru:
this also to align with this dao signer payload: https://github.com/BalancerMaxis/multisig-ops/pull/2664

PR #2681, opened 2026-02-13
Update StableSurge Hook Parameters for JUSD / sJUSD / USDC on Ethereum

tiago-hansen:
This PR decreases the max surge fee from 95.00% to 9.00% and increases the surge threshold from 30.00% to 40.00% for the Aegis JUSD / sJUSD / USDC pool (0xb20fa4) on Ethereum.


Xeonus:
@tiago-hansen can you link the reasoning here? Partner request? Reduction so that aggregators can pick up volume again? Would be great to coordiante this with @zekraken-bot

PR #2754, opened 2026-04-08
feat: notional whitehat merkl reward redirects (eth + arb)

gosuto-inzasheru:
need to confirm all their contract addresses still


Xeonus (approved):
LGTM. "froms" is old contract, "to" is safe Notional provided effectively redirecting claim to their requested target

PR #2762, opened 2026-04-13
feat: redo wpol campaign

gosuto-inzasheru:
recovered amount of wpol of 1454128 was falsely attributed to one pool (0xcd78a20c597e367a4e478a2411ceb790604d7c8f000000000000000000000c22)

this should have been 1249510, with 196228wpol to x951d84e358dd8ad6b3ae6220981bcfc4baf95c84000000000000000000000d62 and 9977wpol to 0x8159462d255c1d24915cb51ec361f700174cd99400000000000000000000075d

this payload recreates the now 'cancelled' old one, but with correct numbers

### DIFF
--- a/BIPs/2026-W18/BIP-892-137_redo_wpol_campaign.json
+++ b/BIPs/2026-W18/BIP-892-137_redo_wpol_campaign.json
@@ -1,0 +1,141 @@
+{
+  "version": "1.0",
+  "chainId": "137",
+  "meta": {
+    "name": "Redo WPOL Campaign - BIP-892",
+    "txBuilderVersion": "1.16.3",
+    "createdFromSafeAddress": "0xeE071f4B516F69a1603dA393CdE8e76C40E5Be85"
+  },
+  "transactions": [
+    {
+      "to": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
+      "value": "0",
+      "data": null,
+      "contractMethod": {
+        "inputs": [
+          {
+            "internalType": "address",
+            "name": "spender",
+            "type": "address"
+          },
+          {
+            "internalType": "uint256",
+            "name": "amount",
+            "type": "uint256"
+          }
+        ],
+        "name": "approve",
+        "payable": false
+      },
+      "contractInputsValues": {
+        "spender": "0x467665D4ae90e7A99c9C9AF785791058426d6eA0",
+        "amount": "5927159439709870321853252"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
underlying recalcs and airdrop file can be found here: https://github.com/maxyz-xyz/balancer-upscale-exploit-shares/compare/fix/redo-wpol-campaign?expand=1

Xeonus (changes requested):
Reviewing changes based on mentioned wpol fix branch [here](https://github.com/maxyz-xyz/balancer-upscale-exploit-shares/tree/fix/redo-wpol-campaign) and assuming that data is now correct

1. the original payloads for BIP-982 distributed WPOL based on one pool set, whereas it should be three pools (MaticX-WMatic, TruMatic-WMatic, stMatic-stable pool). I see those artifacts in the branch now
2. Old campaigns were cancelled by DAO signers, so now we can restart a fresh campaign with remaining amounts
3. Transaction flow is (nearly) the same as with others: Approve WPOL to wrapper, Mint WPOL via wrapper, Approve WPOL to Merkl engine, create campaign
4. I see that the approval is a bit bigger `5927159439709870321853252` vs campaign creation which uses `5927159439709870321853251` WEI, which is fine so it doesn' revert

General parameter check:
- all same as old file, only difference is unix start timestamp to start Apr. 8th (1775817206)
- new total is 5927159.44 WPOL

### Issue on total balances
I see on https://polygonscan.com/address/0xee071f4b516f69a1603da393cde8e76c40e5be85 a total balance of  5,708,472.57 WPOL, campaign tries to create one with a required balance of  5,927,159.44 WPOL? This means the DAO multi-sig doesn't hold enough WPOL to fill the campaign, it's currently short 218,687 WPOL!


The tenderly simulation actually does not revert because the wrapper simply mints the needed tockens but doesn't check if underlying token amounts are present (pull tokens on claim). Claiming would eventually fail if all remaining tokens are claimed before the campaign "completes" with the total target amount of 5.9M WPOL.

So either fix artifacts and total amounts to 5.7M WPOL or top up the 218k mssing WPOL on the multi-sig?

gosuto-inzasheru:
the 219k wpol gap is created by inflated amounts for the old campaign that have already been claimed (we cannot retrieve these)

how that gap will be filled remains to be seen, first prio is to get the campaign back online with correct numbers

Xeonus (approved):
Overall payload generation and artifacts look good. Reasoning for WPOL gap is given. Claim process can continue and has higher priority than dealing with the WPOL gap.

PR #2764, opened 2026-04-13
chore: some folder restructuring for this week's payloads

gosuto-inzasheru:
- move tc/opco/onchainltd payloads out of dao dir
- more dao payloads from week18 to week16


Xeonus (approved):
Cleaning up bundle, moving to correct dirs

PR #2766, opened 2026-04-13
chore: isolate emissions stop payload

gosuto-inzasheru:


gosuto-inzasheru (commented):
this tx needs to have its own nonce and should thus not be batched

Xeonus (approved):
As discussed execute on own nonce between April 16th and before April 23rd, rest executed in separate batch

PR #2768, opened 2026-04-14
Aura Locker: Disable module, cancel upkeep

Xeonus:
- Disable the locker module
- Cancel the chainlink upkeep.

We should also then claim the LINK after this has been done (can be bundled at a later time point)


gosuto-inzasheru (approved):
this is indeed the active `AuraLockerModule`

Image

Xeonus:
Queued

Xeonus:
Executed

PR #2769, opened 2026-04-14
Add Reward to Gauge on Ethereum

marcusblabs:
This PR adds reward (token 0x89bb79 to gauge 0xeec405) on Ethereum.


gosuto-inzasheru:
enables injection of ausdcusdt bpt as a reward token into the gauge for itself (`lp_token` for the gauge is the same bpt) via injector 0x1195

@zekraken-bot do we have a keeper set up for that injector..?

PR #2775, opened 2026-04-23
[BIP-919]: Swap fee change update (add newly created pools)

Xeonus:
Add delta of pools that were created since payload generation. These are 36 pools across 5 networks (verified via api-v3).


gosuto-inzasheru (commented):
i think part2, 8 and 9 for ethereum could be consolidated?

Xeonus:
@gosuto-inzasheru 

- agreed on merging parts 8 and 9 on mainnet, see latest commit
- part 2 needs to stay at 75 pools because this list contains the set with accumulated fees and therefore causes higher total gas because when triggering the fee update, protocol fees are withdrawn. Therefore keep as is to stay on the safe side with the gas limit.
- only other chains with delta present are ARB, BASE, GNO, Plasma. Previous changes to other chains like HyperEVM, AVAX, XLayer and Monad have already been approved. No changes on those detected. Fee change should be global independet of the chain designated as "core" or not.

gosuto-inzasheru (approved):

PR #2778, opened 2026-04-29
feat: setup raw fee collection on monad and hyperevm

gosuto-inzasheru:
allow `ProtocolFeeSweeperV2` to withdraw protocol fees and send raw fee tokens to the v3 fee bot eoa for manual processing


raq0x (approved):
👍

PR #2781, opened 2026-05-01
feat: redir notional beneficiary on arb1

gosuto-inzasheru:


Xeonus (approved):
New target address is a match mentioned here

Image

froms also confirmed via chat logs.

No labels added in `bal_addresses` book though.

PR #2784, opened 2026-05-04
[BIP-920] veBAL Compensation Airdrop

gosuto-inzasheru:
as per https://forum.balancer.fi/t/bip-xxx-vebal-compensation-airdrop/7025

addresses for top 3 receivers have been adjusted as requested (see forum thread)


Xeonus (approved):
- Matches prop and dune data
- BIP ID OK

PR #2791, opened 2026-05-12
[BIP-873] Treasury -> OpCo USDC Transfer

Xeonus:
As per BIP-873, provide last tranche of 717531 USDC to OpCo


gosuto-inzasheru (dismissed):
amount is correct, treasury safe just needs usdc before this can be executed successfully

gosuto-inzasheru (approved):

PR #2806, opened 2026-05-26
fix: process OmniAutoOps payload reports

gosuto-inzasheru:
follow-up to #2805


Xeonus (approved):
Thanks, I missed those!

PR #2839, opened 2026-06-24
Changes gas station's keeper address to Tenderly signer

raq0x:
As a part of migration from Chainlink automation, we will handle the upkeeps from Tenderly actions and the keeper address needs to be changed to Tenderly action's signer


gosuto-inzasheru (approved):
lgtm

gosuto-inzasheru:
@raq0x / @Xeonus this was not placed in the `OmniAutoOps` dir, so did not get queued automatically

PR #2845, opened 2026-06-30
Set allow-unsafe-pr-checkout on fork PR report checkout

gosuto-inzasheru:
actions/checkout@v7 refuses to check out fork PR code from a pull_request_target context by default. The run_reports reusable workflow intentionally checks out fork head code (secrets are gated behind the 'external' environment), so opt in explicitly.


Xeonus (approved):
Matches implementation approach from warning message from https://github.com/balancer/multisig-ops/actions/runs/28445552311/job/84293873753

PR #2851, opened 2026-07-10
feat: withdraw all chainlink upkeeps

gosuto-inzasheru:
needs 50 blocks passed after exec of #2849


Xeonus (approved):
- all withdrawals send LINK to `gosuto_deployer`
- confirmed no upkeeps are missed and validated against `ops.balancer.fi/api/injector/v2/all`
- zkEVM is shut-down now although one injector was there, no real loss though

PR #2859, opened 2026-07-31
feat: claim and sweep all $safe and $cow to treasury

gosuto-inzasheru:


gosuto-inzasheru:
stage2 sims are failing, these need to be execd at least 24 hours after stage1 (dao, opco)

